PR #17, opened 2021-11-01
Fix lexing of integer literals with leading zero

g4rgoil:


run
```diff
@@ -17,4 +17,4 @@ EXE_PATH=$(find $SCRIPT_PATH/target/java-compiler-*-with-dependencies.jar) || ex
 
 # execute the compiler with the provided arguments and return the exit code
 java -jar "$EXE_PATH" "$@"
-exit $?
+exit $?
```
larsk21, line 20: Is this part of fixing the lexing of integer literals?

src/test/java/edu/kit/compiler/lexer/LexerTest.java
```diff
@@ -48,6 +48,11 @@ public void testUnicode() throws LexException {
         var lexer = new Lexer(getIterator("int \u309e = 1;"));
         assertEquals(Keyword_Int, lexer.getNextToken().getType());
         assertThrows(LexException.class, () -> lexer.getNextToken());
+        var lexer2 = new Lexer(getIterator("int 😀 = 1;"));
```
larsk21, line 51: Same here, shouldn't this be in its own PR?

src/test/java/edu/kit/compiler/lexer/LexerTest.java
```diff
@@ -168,7 +173,8 @@ public void testZeroLiteral() throws LexException {
     @Test
     public void testIllegalLiteral() throws LexException {
```
larsk21, line 174: This is not illegal anymore, please rename this test.

larsk21 (changes requested):

run
g4rgoil: No, it's an unrelated commit I hadn't noticed.

src/test/java/edu/kit/compiler/lexer/LexerTest.java
g4rgoil: This one didn't make it into the original lexer PR. It's technically unrelated to this fix.

src/test/java/edu/kit/compiler/lexer/LexerTest.java
```diff
@@ -166,9 +171,10 @@ public void testZeroLiteral() throws LexException {
     }
 
     @Test
-    public void testIllegalLiteral() throws LexException {
+    public void testLeadingZero() throws LexException {
```
larsk21, line 174: Just one last thing, could you also remove the `throws` declaration, now that there's nothing that should be thrown anymore? 😅

g4rgoil: For the test to compile, it still needs a `throws` declaration, because we're using checked exceptions in the lexer. The test would still fail if an exception we're to be thrown.

larsk21 (approved):

PR #25, opened 2021-11-04
Disable partial matching in command line parser

### LINKED ISSUE #20: CLI allowing command prefixes
Currently our CLI allows to use for example `--lext` instead of `--lextest`. We should check if there is any easy way to disable this in the library we're using (Apache Commons CLI).

g4rgoil:
Resolves #20


N-Maas (approved):

larsk21:
The new version of the CLI requires a non-empty usage string for the HelpFormatter.

PR #27, opened 2021-11-05
Proposal for logging system

g4rgoil:
This PR proposes a unified logging system for our compiler. Please let me know if you would like changes to the overall design or to the API of `Logger`. Console output is currently structured as follows:

```
error: lexer: 42,28: unexpected character '$'
```

### DIFF
--- a/run
+++ b/run
@@ -15,6 +15,11 @@ popd > /dev/null
 # find the compiler executable (EXE_PATH)
 EXE_PATH=$(find $SCRIPT_PATH/target/java-compiler-*-with-dependencies.jar) || exit $?
 
+# enable colored logging if stderr is a terminal
+if [[ -t 2 ]]; then
+    export COLOR="on"
+fi
+
 # execute the compiler with the provided arguments and return the exit code
 java -jar "$EXE_PATH" "$@"
 exit $?
--- a/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
+++ b/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
@@ -22,8 +22,6 @@ import java.io.OutputStream;
 
 @Slf4j
 public class JavaEasyCompiler {
-    private Logger logger = new Logger();
-
     /**
      * Output the file contents to stdout.
      * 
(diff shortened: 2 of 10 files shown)


0xNB:
I propose that we change out the logger here for a Static Logger from `logback`. We can set the level programatically without the need to pass it from function call to function call.

g4rgoil:
> I propose that we change out the logger here for a Static Logger from `logback`. We can set the level programatically without the need to pass it from function call to function call.

From what I can tell, `logback` is intended more for logging to proper, and especially filterable log files (e.g. by default it includes timestamp and class in the output).  During the meeting on Wednesday, we agreed on a solution similar to what is proposed in this PR.

N-Maas:
> error: lexer: 42,28: unexpected character '$'

This is a bit weird to read due to the multiple colons and extreme terseness. I would prefer something like

`error at lexer: line 42, column 28: unexpected character '$'`

or

`error during lexing at line 42, column 28: unexpected character '$'`

g4rgoil:
> `error at lexer: line 42, column 28: unexpected character '$'

Would you be okay with the following? This would work better for different log levels (e.g. `info at lexer: ...` feels a bit strange).

```
error: lexer at line 42, column 28: unexpected character '$'
```

N-Maas:
> > `error at lexer: line 42, column 28: unexpected character '$'
> 
> Would you be okay with the following? This would work better for different log levels (e.g. `info at lexer: ...` feels a bit strange).
> 
> ```
> error: lexer at line 42, column 28: unexpected character '$'
> ```

Seems reasonable to me

0xNB:
I don't find any large prolems with this PR except for some things like different tastes of code style which do not matter. However I find coding the return code of our Compiler into the `CompilerException` a strange design choice, but since the `CompilerException` should only be used in the main class its OK.

g4rgoil:
> I don't find any large prolems with this PR except for some things like different tastes of code style which do not matter. However I think that implementing the return code of our Compiler into the `CompilerException` is a strange design choice. But since the `CompilerException` should only be used in the main class its OK.

We don't really have defined code style for this project (yet?), so I just do what I believe is acceptable.
I agree with your second point. The problem is, that we need some way of getting from compiler exceptions to the exit codes. I just reused the `Result` type defined in our main class for this purpose. There are definitely better ways of going about this. But as you pointed out, we only use the exit codes in a single place, so I think it's fine for now.

N-Maas (approved):

PR #28, opened 2021-11-05
cmdfix line endings

0xNB:


larsk21:
Could you elaborate what this does? I thought we were supposed to output the file *exactly* the way we read it. Why are you replacing line endings again?

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -102,8 +116,15 @@ public static void main(String[] args) {
             result = Result.Ok;
         } else if (cmd.hasOption("e")) {
             String filePath = cmd.getOptionValue("e");
-
-            result = echo(filePath);
+            try {
+                String content = echo(filePath);
+                System.out.println(content);
```
larsk21, line 121: Why do we print the string here and not in `echo`? I would prefer keeping the logic/action in the separate methods and using the main method only for command line parsing and stuff like that.

0xNB: For testing purposes

PR #37, opened 2021-11-09
Symboltable

0xNB:


src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Map;
+
+public class NamespaceMapper {
+
+    private final Map<ClassNode, ClassNamespace> symbolTableMap = new HashMap<>();
+
+    @AllArgsConstructor
+    @Data
+    public static class ClassNamespace {
+        private SymbolTable methodSymbols;
```
g4rgoil, line 17: In the AST we differentiate between static methods and dynamic methods, would it make sense to make that distinction here as well?

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
```diff
@@ -0,0 +1,42 @@
...
+        private SymbolTable fieldSymbols;
```
g4rgoil, line 18: Do we need a full fledged `SymbolTable` here? This is a simple mapping between an identifier and a field or method respectively. Would an ordinary `Map` not do the job just as well?

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
```diff
@@ -0,0 +1,143 @@
...
+
+@Data
+public class SymbolTable  {
+
+    private final ArrayDeque<Change> changes = new ArrayDeque<>();
+    private final SymbolStringTable symbolStringTable = new SymbolStringTable();
+    private Scope currentScope;
+
+    @RequiredArgsConstructor
+    static class Symbol {
```
g4rgoil, line 19: You're nested classes should probably be either explicitly `private` or `public` depending on whether they are implementation detail or public interface.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
```diff
@@ -0,0 +1,143 @@
...
+            }
+            c.sym.currentDefinition = c.previousDefinition;
+            c.sym.currentScope = c.previousScope;
+        }
+        currentScope = currentScope.parent;
+    }
+    public Symbol insert(int name, Definition definition) {
+        Symbol s = symbolStringTable.find(name);
+        if(isDefinedInCurrentScope(s)) {
+            // this should never happen, make sure to check in upper level to get line number and column
```
g4rgoil, line 98: What do mean by "this should never happen" here? Correct me if I'm wrong, but this condition occurs, if a variable is declared twice in the same scope. That is definitely an error, but can (and will) absolutely happen.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
N-Maas: We should definitively differentiate the two. The question is probably whether we want to add any static method to the name space or, instead, directly check whether it is the main method (which is the only semantically valid static method).

larsk21:
Also, I would prefer to merge #38 first, then rebase this PR and merge it directly into master.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
```diff
@@ -0,0 +1,143 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+
+import java.util.ArrayDeque;
+import java.util.LinkedHashMap;
+
+@Data
+public class SymbolTable  {
```
larsk21, line 12: As far as I can see, `currentScope` is not set during the construction of `SymbolTable`, right? So we have to call `enterScope` at least once before using the `SymbolTable` for insertion/lookup? If that's the case, we should probably add documentation about it to avoid null pointer exceptions.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private final ArrayDeque<Change> changes = new ArrayDeque<>();
+    private final SymbolStringTable symbolStringTable = new SymbolStringTable();
+    private Scope currentScope;
+
+    @RequiredArgsConstructor
+    static class Symbol {
+        @NonNull
+        int name;
```
larsk21, line 21: `@NonNull` is unnecessary on primitive type variables.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
```diff
@@ -0,0 +1,143 @@
...
+            if (c.previousScope == null) {
+                symbolStringTable.remove(c.sym.name);
+                return;
+            }
+            c.sym.currentDefinition = c.previousDefinition;
+            c.sym.currentScope = c.previousScope;
+        }
+        currentScope = currentScope.parent;
+    }
+    public Symbol insert(int name, Definition definition) {
```
larsk21, line 95: Why is the additional `name` parameter necessary? Is there a scenario where we want to insert a definition with a name different from `definition.getName()`?

larsk21:
`MethodNodeParameter` should implement `Definition` as well.

src/main/java/edu/kit/compiler/semantic/SemanticException.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Optional;
+
+public class SemanticException extends CompilerException {
+
+    public SemanticException(String msg) {
+        super(msg);
+    }
+
+    @Override
+    public Optional<SourceLocation> getSourceLocation() {
```
larsk21, line 15: Please add a constructor for passing a `SourceLocation` with this exception.

g4rgoil:
> `MethodNodeParameter` should implement `Definition` as well.

As long as fields are stored in a `SymbolTable`, `ClassNodeField` needs to implement `Definition` as well.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
larsk21: I actually find this behavior useful, but we should definitely document it.

larsk21:
> > `MethodNodeParameter` should implement `Definition` as well.
> 
> As long as fields are stored in a `SymbolTable`, `ClassNodeField` needs to implement `Definition` as well.

Yes and also to make them referenceable (`Reference` interface will provide `getDefinition() : Definition`).

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
```diff
@@ -0,0 +1,42 @@
...
+import edu.kit.compiler.data.ast_nodes.ClassNode;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class NamespaceMapper {
+
+    private final Map<ClassNode, ClassNamespace> symbolTableMap = new HashMap<>();
```
larsk21, line 12: Could we also map from identifier (i.e. `int`) to `ClassNamespace` and include a `ClassNode` attribute in `ClassNamespace`? That way you could easily map from a `DataType` to the corresponding class and access its attributes/fields.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
larsk21: Also, we could map to `MethodNode`/`ClassNodeField` directly to gain access to e.g. method argument types without having to cast the returned `Definition`.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
0xNB: Yes this is definitely an error but this shold not be checked by the symbol table itself but the code surrounding it.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
0xNB: Yes the idea is that every variable has to be defined inside a scope. They get removed from the symbol table if you leave the scope that defined it. In code this is done by checking for `null` refernce to the current definition on `leaveScope()`

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
0xNB: This is used to let lombok generate the `@RequiredArgsConstructor`

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
0xNB: Right now I can't think of one. I mean we could leave it out and when we come across a case where this is necessary we add it again.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
0xNB: Yes, i remember that we discussed this earlier for the symbol table as well. I think its good to just stick to one solution, therefore we should only reference our symbols and classes with their respective id not with the data type itself.

larsk21: It's also relevant when you have, for example, a `Definition` of a variable with a `DataType` and want to access the list of methods available on that variable.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
larsk21: This leaves a warning in my IDE, couldn't we make `name` final?

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
larsk21: Why I asked to include `ClassNode` is, if you have an "active" namespace (i.e. a variable `currentNamespace : ClassNamespace` and encounter a `this` pointer you have to find the class data type, which is (for now) not stored in the namespace.

larsk21:
Please remember to rebase the branch to only list the actually relevant commits.

src/main/java/edu/kit/compiler/semantic/SemanticException.java
```diff
@@ -7,8 +7,22 @@
 
 public class SemanticException extends CompilerException {
 
+    private final SourceLocation sourceLocation;
+
+    /**
+     * This constructor should be avoided. Implementation needs to switch to the one with source location once we
+     * get the error location in our AST.
```
larsk21, line 14: The error location is already in our AST, isn't it? `AstNode` has `line` and `column` as fields.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
0xNB: I would just use the extra data structure even if it costs us a tiny amount of performance, since its nicer to read

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
0xNB: @g4rgoil 
Yes I think we can shorten the definition here. The mapper is only used for looking up if a method or field is defined anyway and give the correct. So the mapper holds one symbol table per class and defined methods and fields

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final Map<Integer, ClassNamespace> symbolTableMap = new HashMap<>();
+
+    @Data
+    public static class ClassNamespace {
+        private final int classNodeRef;
+        private final Map<Integer, MethodNode.DynamicMethodNode> dynamicMethods = new HashMap<>();
+        private final Map<Integer, MethodNode.StaticMethodNode> staticMethods = new HashMap<>();
+        // Inner symbol table that holds most recent definitions of control flow
+        private final SymbolTable classSymbolTable = new SymbolTable();
```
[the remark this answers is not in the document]
0xNB: Getters and Setters are defined by the `@Data` Annotation

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
```diff
@@ -0,0 +1,64 @@
...
+     * Create a new symbol table once a new class is visited during recursive descent.
+     * @param classNode The classnode for which the new namespaces are created.
+     * @return classNamespace The class namespace which contains a reference to the class node and methods/fields
+     */
+    public ClassNamespace insertSymbolTable(ClassNode classNode) {
+        int classNodeName = classNode.getName();
+        if(symbolTableMap.containsKey(classNodeName)) {
+            // this should never happend make sure to fix all issues where the same class would get added twice
+            CompilerException.SourceLocation sl = new CompilerException.SourceLocation(classNode.getLine(), classNode.getColumn());
+            throw new SemanticException("Cannot insert same class twice into namespace map", sl);
```
g4rgoil, line 39: I've now had to write pretty much the same piece of code as well. This is a lot of repetition. What do you think about turning `SourceLocation` into an interface and implement it wherever we have `getLine` and `getColumn`?

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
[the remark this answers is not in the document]
g4rgoil: My bad, forgot about Lombok.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class NamespaceMapper {
+
+    private final Map<Integer, ClassNamespace> symbolTableMap = new HashMap<>();
+
+    @Data
+    public static class ClassNamespace {
+        private final int classNodeRef;
```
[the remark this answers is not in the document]
larsk21: Second this, see my other comment for the arguments.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
larsk21: Already opened an issue yesterday. We don't have to make it an interface, but should use it instead of line/column where applicable.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
larsk21, line 21: What is this `classSymbolTable` for? Given I use the `getClassNamespace` method, what will I find in this symbol table? The temporary symbol table when going through methods is a already a field in my visitor class, if that was the intention.

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
larsk21: Though I would propose to open a new PR for this and close this one as fast as possible.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
```diff
@@ -0,0 +1,156 @@
...
+        Symbol s = symbolStringTable.find(symbol);
+        if(symbolStringTable.find(symbol) == null) {
+            return null;
+        }
+        else {
+            return s.currentDefinition;
+        }
+    }
+
+    public boolean isDefinedInCurrentScope(Symbol symbol) {
```
larsk21, line 147: Shouldn't this take an `int` as argument? What I want to do is check if my local variable is defined in the current scope, having its name. If not, where would I get the symbol from, in order to call this method?

src/main/java/edu/kit/compiler/semantic/NamespaceMapper.java
0xNB: This would be the per class symbol table but if you already defined this in your Visitor it should be als okay then I remove it here.

src/main/java/edu/kit/compiler/semantic/SymbolTable.java
0xNB: Changed it to take an integer that represents the symbols name

larsk21:
This suddenly includes a lot of changes to `Parser`. I suppose this is by accident?

larsk21 (approved):

PR #38, opened 2021-11-10
Ast construction

N-Maas:


src/main/java/edu/kit/compiler/lexer/StringTable.java
```diff
@@ -33,7 +33,7 @@ public String retrieve(int hash) {
         if(hash < size() && hash >= 0) {
             return strings.get(hash);
         }
-        return null;
+        throw new IllegalArgumentException("No string present for this value.");
```
larsk21, line 36: Valid behavior, but please change the doc accordingly. Also we should probably add a `contains` method to safely access strings.

src/main/java/edu/kit/compiler/parser/Parser.java
```diff
@@ -40,81 +74,117 @@ public Parser(Lexer lexer) {
...
+                firstToken.getLine(), firstToken.getColumn(), name.getIntValue().get()
+            ));
             expect(TokenType.Operator_BraceR);
         }
         expect(TokenType.EndOfStream);
+        return new ProgramNode(programStart.getLine(), programStart.getColumn(), classes, false);
     }
 
-    private void parseClassMembers() {
+    private ClassNode parseClassMembers(int line, int column, int className) {
```
larsk21, line 94: This method mixes parsing the class members and creating the actual class node. From the name alone, I wouldn't expect it to return an entire class node and also wouldn't understand what the `line`, `column` and `className` parameters are for.

Would it be a solution to split this into `parseClass` and maybe `parseStaticMethod`, `parseDynamicMethod` and `parseField`?

src/main/java/edu/kit/compiler/parser/Parser.java
```diff
@@ -123,160 +193,201 @@ private void parseBlock() {
...
                 }
                 expect(TokenType.Operator_Semicolon);
+                stmts.add(new LocalVariableDeclarationStatementNode(
+                    firstToken.getLine(), firstToken.getColumn(), variableType, name.getIntValue().get(), expr, false
+                ));
             } else {
                 // Statement
-                parseStatement();
+                Optional<StatementNode> stmt = parseStatement();
+                stmt.ifPresent(node -> stmts.add(node));
```
larsk21, line 211: How fast are lambdas in Java? Seems like it would be simple to write it without here.

src/main/java/edu/kit/compiler/parser/Parser.java
```diff
@@ -123,160 +193,201 @@ private void parseBlock() {
...
-                parseExpression();
+                ExpressionNode condition = parseExpression();
                 expect(TokenType.Operator_ParenR);
-                parseStatement();
+                StatementNode thenStmt = parseStatement().orElse(emptyBlock(token));
+                Optional<StatementNode> elseStmt = Optional.empty();
                 if (tokenStream.get().getType() == TokenType.Keyword_Else) {
                     tokenStream.next();
-                    parseStatement();
+                    elseStmt = Optional.of(parseStatement().orElse(emptyBlock(token)));
```
larsk21, line 238: That's a lot of options around here. Would it be possible to simply return a `StatementNode` from `parseStatement` (with `emptyBlock` instead of `Optional.empty`? This way we would also keep empty blocks inside another block in the statement list, which could be required for the pretty printer(?).

src/test/java/edu/kit/compiler/parser/ParserTest.java
```diff
@@ -52,6 +57,127 @@ public void testInvalidIdentifier() {
         assertThrows(ParseException.class, (parser)::parse, "not part of language spec");
     }
 
+    @Test
+    public void testSimpleAst() {
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        PrintStream sysout = System.out;
+        System.setOut(new PrintStream(stream));
```
larsk21, line 64: Why not put this in a `BeforeEach`/`AfterEach`?

larsk21 (commented):
Looks good to me. Did you want to add some MJ tests in this PR already?

src/main/java/edu/kit/compiler/parser/Parser.java
N-Maas: I did return a `StatementNode` at first, but the pretty print example in the assignment explicitly deletes empty statements, so I decided to implement the same behavior. It was also mentioned in the meeting that we can delete empty statements.

src/main/java/edu/kit/compiler/parser/Parser.java
N-Maas: `parseDynamicMethod` and `parseField` are a bit hard to separate, but `parseStaticMethod` is separate now and the function parses the whole class.

N-Maas:
> Did you want to add some MJ tests in this PR already?

I wanted to add some, but as far as I can tell, mjtest only checks the return code of the pretty print call. To me, this doesn't seem to provide any benefit over the already existing parser tests (which are automatically included in the ast tests). But I added the ast test call to the CI now.

src/main/java/edu/kit/compiler/parser/Parser.java
larsk21: I see. You could catch empty statements in the `parseBlock` method like local variable declarations. But I don't know if that's more readable.

N-Maas: Good idea, I think this improves it a bit

larsk21 (approved):

PR #40, opened 2021-11-10
add AST object

larsk21:
- add AstObject class
- make AstNode and several AST classes extend AstObject
- pass position to new AstObject classes during parsing


src/main/java/edu/kit/compiler/parser/Parser.java
```diff
@@ -37,17 +37,12 @@
...
-import java.lang.StackWalker.Option;
-import java.lang.Thread.State;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
-import javax.xml.crypto.Data;
-
```
N-Maas, line 50: Good catch :sweat_smile:

N-Maas (approved):

PR #41, opened 2021-11-11
Assignment 4 - Abstract syntax and attribute grammar

g4rgoil:
This is a first (mostly complete) version of Assignment 4 (~[assignment-04.pdf](https://github.com/larsk21/compiler-minijava/files/7528384/assignment-04.pdf)~).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -374,4 +374,4 @@ TSWLatexianTemp*
 # REVTeX puts footnotes in the bibliography by default, unless the nofootinbib
 # option is specified. Footnotes are the stored in a file with suffix Notes.bib.
 # Uncomment the next line to have this generated file ignored.
-#*Notes.bib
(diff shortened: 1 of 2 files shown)


larsk21:
I just quickly looked over it, but it seems alright. You used attributes like `.node` to pass the created node to the parent node, except for `ClassDeclaration`, which is used directly. Is there a reason for this?

g4rgoil:
> I just quickly looked over it, but it seems alright. You used attributes like `.node` to pass the created node to the parent node, except for `ClassDeclaration`, which is used directly. Is there a reason for this?

It's not like this on purpose, thanks for pointing that out to me. It's simply the first rule I wrote, after which I pretty much just forgot about it.

larsk21 (approved):

PR #44, opened 2021-11-15
Detailed Name and Type Analysis

larsk21:


src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -0,0 +1,491 @@
...
+        return resultType;
+    }
+
+    public DataType visit(ValueExpressionNode valueExpressionNode) {
+        DataType resultType;
+        switch (valueExpressionNode.getValueType()) {
+        case False:
+            resultType = new DataType(DataTypeClass.Boolean);
+            break;
+        case IntegerLiteral:
```
g4rgoil, line 436: We need to ensure that integer literals actually fit into a 32-bit signed integer. So you need something like the following:
```
try {
    valueExpressionNode.literalValue.get().intValue();
} catch (NumberFormatException e) {
    // ...
}
```

larsk21: Good point, I thought we already checked that elsewhere.

larsk21:
I removed the use of `ErrorHandler` temporarily, because not stopping in case of an error leads to various side effects in the visitor (for example: an invalid identifier can still be used in a field access, but the namespace of the identifier type might not be available).

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -0,0 +1,505 @@
...
+ * - void is not used as a field, parameter or local variable type, in a new
+ * object or new array expression
+ * - hasError is set in all nodes where an error occured
+ * - user defined data types of fields and methods (incl. parameters) are valid
+ * - integer literal values are valid integer values (32-bit signed)
+ * 
+ * Not checked:
+ * - all code paths return a value
+ * - left side of an assignment is an l-value
+ */
```
g4rgoil, line 41: Is the check for whether `args` is used in a main method responsibility of the third pass?

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -0,0 +1,505 @@
...
+            }
+        } else if (expectedReturnType.getType() != DataTypeClass.Void) {
+            semanticError(returnStatementNode, "method requires a return value of type %s", expectedReturnType.getRepresentation(stringTable));
+        }
+
+        return VoidType;
+    }
+
+    public DataType visit(ExpressionStatementNode expressionStatementNode) {
+        expressionStatementNode.getExpression().accept(this);
```
g4rgoil, line 258: Only method invocations and assignments are permitted in _ExpressionStatements_, should this be checked in your phase?

> _ExpressionStatements_, die keine Methodenaufrufe oder Zuweisungen sind, müssen
in der Semantikphase abgewiesen werden. Insbesondere haben _NewObjectExpressions_
in MiniJava keine Seiteneffekte und werden daher ebenfalls abgewiesen.

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -0,0 +1,505 @@
...
+            }
+
+            if (!parameter.isHasError()) {
+                symboltable.insert(parameter);
+            }
+        }
+
+        expectedReturnType = methodNode.getType();
+        methodNode.getStatementBlock().accept(this);
+        expectedReturnType = null;
```
g4rgoil, line 155: You could get rid of this statefulness by adding an inner class which extends your visitor and has an extra field for the expected return type. There is only a single method that relies on this information, so you would only have to move that one into the inner class.

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -0,0 +1,505 @@
...
+            break;
+        default:
+            throw new IllegalArgumentException("unsupported value expression type");
+        }
+
+        valueExpressionNode.setResultType(resultType);
+        return resultType;
+    }
+
+    public DataType visit(NewObjectExpressionNode newObjectExpressionNode) {
```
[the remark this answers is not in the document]
N-Maas: This is my job (third pass)

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
N-Maas: Also my job

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
N-Maas: Also my job :D

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
N-Maas: If we want to avoid state, I would rather suggest to not use the visitor for iterating over the methods at all and instead do this in a separate function, creating a new visitor for each method. This is also what I'm doing for the third pass.

PR #45, opened 2021-11-16
Definition and positionable

N-Maas:


src/main/java/edu/kit/compiler/semantic/SemanticException.java
```diff
@@ -28,4 +37,12 @@ public SemanticException(String msg, SourceLocation sourceLocation) {
     public JavaEasyCompiler.Result getResult() {
         return JavaEasyCompiler.Result.SemanticError;
     }
+
+    private static String errorListToMsg(Iterable<SemanticError> errors) {
```
g4rgoil, line 41: Why do we need to accumulate the errors? Couldn't we just dump each one into a logger as we go?

N-Maas: Do you mean dumping the error when encountering it, or dumping it when constructing the SemanticException? The latter would add a significant side effect to the constructor, which seems rather weird to me

g4rgoil: I mean dumping the errors as soon as we encounter them. If I see it right, your intention is to have a collection of `Errors` in your class, which you would turn into a single exception at the end of the pass. I would propose some kind of message queue, with a method like `void receive(Error error)`. The implementation of that queue could then for example pass the errors on to a logger, or store them if desired.

g4rgoil: The reason why I would prefer not to wrap all errors into a single exception is that it would allow us to keep our log output consistent.

N-Maas: Okay, I have added a queue as proposed.

Note that the queue has an `hasError`-flag, so we can check later on whether any error occured.

src/main/java/edu/kit/compiler/data/CompilerException.java
```diff
@@ -21,7 +19,7 @@ public CompilerException(String message) {
     /**
      * Return the optional source location of the character that caused the exception.
      */
-    public abstract Optional<SourceLocation> getSourceLocation();
+    public abstract Optional<Positionable> getSourceLocation();
```
0xNB, line 22: `getPosition()` instead of `getSourceLocation()`

src/main/java/edu/kit/compiler/semantic/ErrorQueue.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Getter;
+import lombok.NonNull;
+
+public class ErrorQueue {
+    @Getter
+    private boolean hasError;
+    @NonNull
+    private Logger logger;
+
+    public ErrorQueue(Logger logger) {
```
0xNB, line 13: Why isn't this queue a queue? Right now its just an object that logs a source location?

N-Maas: Well, I can rename it to `ErrorHandler` or something

g4rgoil (approved):

PR #46, opened 2021-11-16
Semantic Analysis - Namespace Gathering

g4rgoil:


N-Maas (approved):
Seems generally okay to me. If you include the error handling from #45, I would also suggest to merge this PR quickly afterwards so that we can build upon it. We can do some cleanup later.

N-Maas:
Hint: you can run the tests locally via `./run_mjtests.sh {lexer, syntax, ast, semantic, ...}`

(you might first need `git submodule init`)

g4rgoil:
> Hint: you can run the tests locally via `./run_mjtests.sh {lexer, syntax, ast, semantic, ...}`
> 
> (you might first need `git submodule init`)

Yes, I have that set up. Unfortunately I tend to forget about it :D

PR #48, opened 2021-11-17
Semantic other checks

N-Maas:
Third round of semantic checks

### DIFF
--- a/src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
+++ b/src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
@@ -1,0 +1,169 @@
+package edu.kit.compiler.semantic;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.ArrayList;
+
+import edu.kit.compiler.data.AstVisitor;
+import edu.kit.compiler.data.ast_nodes.ClassNode;
+import edu.kit.compiler.data.ast_nodes.ExpressionNode;
+import edu.kit.compiler.data.ast_nodes.MethodNode;
+import edu.kit.compiler.data.ast_nodes.ProgramNode;
+import edu.kit.compiler.data.ast_nodes.StatementNode;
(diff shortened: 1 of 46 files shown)


src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
```diff
@@ -0,0 +1,253 @@
...
+                boolean returns = method.getStatementBlock().accept(visitor);
+                if (!returns && !method.getType().equals(voidType)) {
+                    errorHandler.receive(new SemanticError(method,
+                        "return statement for all branches required"));
+                }
+            }
+            for (MethodNode method: currentClass.getStaticMethods()) {
+                MethodCheckVisitor visitor = new MethodCheckVisitor(true, errorHandler, stringClass);
+                method.getStatementBlock().accept(visitor);
+            }
```
g4rgoil, line 62: You could consider checking for `return` on all paths in static methods as well.

src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public Boolean visit(ValueExpressionNode expr) {
+        // nothing to check here
+        return false;
+    }
+
+    public Boolean visit(ThisExpressionNode expr) {
+        if (this.isMain) {
+            errorHandler.receive(new SemanticError(expr,
```
larsk21, line 201: Already checked in second phase

src/main/java/edu/kit/compiler/semantic/SemanticException.java
```diff
@@ -32,4 +45,8 @@ public SemanticException(SemanticError error) {
     public JavaEasyCompiler.Result getResult() {
         return JavaEasyCompiler.Result.SemanticError;
     }
+
+    public boolean surpressMessage() {
```
larsk21, line 49: From the name, I would expect that `supressMessage` sets `quiet` to true.

larsk21 (approved):
Failing tests are not caused by third phase.

PR #52, opened 2021-11-18
Fix some type analysis issues

larsk21:
- local variable can be referenced in its own expression
- Any type is only compatible to reference types
- static methods cannot reference fields (stupid bug, I know)


N-Maas (approved):
If you remove/revert the last commit, I think we can merge this

PR #53, opened 2021-11-18
fix: leaveScope can't handle multiple definitions

larsk21:


g4rgoil (approved):
Now that is a very unfortunate bug.

PR #61, opened 2021-11-22
Fix subtle bug in lexing of comments

g4rgoil:
The new lexer introduced a bug, where closing comments delimiters would be ignored if they were preceded by an asterisk (as in `/***/`).


larsk21 (approved):
Looks good to me.

PR #63, opened 2021-11-23
Remove ReaderCharIterator

larsk21:
Remove `ReaderCharIterator` since it is no longer required since #43.


g4rgoil (approved):
Good point, seems I forgot to remove this class.

PR #64, opened 2021-11-23
Clean up some tests

larsk21:
- rename test classes to fit Maven's naming conventions
- resolve IDE warnings (unused variables)

### DIFF
--- a/src/test/java/edu/kit/compiler/cmd/EchoLineEndingTest.java
+++ b/src/test/java/edu/kit/compiler/cmd/EchoLineEndingTest.java
@@ -11,7 +11,7 @@ import java.io.IOException;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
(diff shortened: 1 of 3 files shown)


0xNB:
Approved, if you want you can add more cleaning up of warnings from our IDEs in this merge request. Or if you want me to I'll collect some

0xNB:
For example we could also think about cleaning up Optionals as return type or inner fields, or remove unused methods

0xNB:
For example we could also think about cleaning up Optionals as return type or inner fields, or remove unused methods

larsk21:
> For example we could also think about cleaning up Optionals as return type or inner fields, or remove unused methods

That would be valid in some places (for example in `DataType`), where context makes it clear whether or not to expect a value. When the caller would have to check for `null` to be safe, we should keep using `Optional` and `isPresent`/`isEmpty` instead.

PR #65, opened 2021-11-24
Remodel standard library

larsk21:
Standard library calls are now a special construct without `System` or `System.out`/`System.in` being defined on its own. The following code yields "`System` cannot be resolved":
```
// inside some method
if (System.out == null) {
  ...
```

detailed changes:
- remove global variable `System`
- change `StandardLibrary` to include only names and method definitions
- explicitly check method invocations to identify standard library calls
- allow removing the object of a method invocation
- add tests for standard library calls

### DIFF
--- a/src/main/java/edu/kit/compiler/data/AstVisitor.java
+++ b/src/main/java/edu/kit/compiler/data/AstVisitor.java
@@ -16,7 +16,6 @@ public interface AstVisitor<T> {
     default T visit(ClassNode classNode) { throw new UnsupportedOperationException(); }
     default T visit(StaticMethodNode staticMethodNode) { throw new UnsupportedOperationException(); }
     default T visit(DynamicMethodNode dynamicMethodNode) { throw new UnsupportedOperationException(); }
-    default T visit(StandardLibraryMethodNode standardLibraryMethodNode) { throw new UnsupportedOperationException(); }
     default T visit(BlockStatementNode blockStatementNode) { throw new UnsupportedOperationException(); }
     default T visit(LocalVariableDeclarationStatementNode localVariableDeclarationStatementNode) { throw new UnsupportedOperationException(); }
     default T visit(IfStatementNode ifStatementNode) { throw new UnsupportedOperationException(); }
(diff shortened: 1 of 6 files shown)


0xNB:
Hmm what about the comparison of `System.in` == `System.in`. Since both return the same stream the check should also work

larsk21:
> Hmm what about the comparison of `System.in` == `System.in`. Since both return the same stream the check should also work

There is no concept of stream in MiniJava. We were told that system calls are special constructs and that comparisons like `System == System` are not valid. I would guess the same holds for `System.in` and `System.out`.

larsk21:
> The implementation in the visitor looks good, it seems to solve the issues of the previous implementation. The only potential problem I see is how we can map the `MethodNode`s of the standard library to the corresponding Firm entities (see [StandardLibraryEntities](https://github.com/larsk21/compiler-minijava/blob/1552174ab99f18ad3bf9de96e09a13e5150759d9/src/main/java/edu/kit/compiler/transform/StandardLibraryEntities.java)).

When reaching a `MethodInvocationExpressionNode`, we would have to check whether it is a standard library call or not. If it is, we can use the `method : StandardLibraryMethod` attribute of `StandardLibraryMethodNode` to determine the correct method.

If you don't want to make this test with `invocation.getDefinition() instanceof StandardLibraryMethodNode` (or a small visitor), we can add a flag to `MethodNode` or something like that.

g4rgoil:
> When reaching a `MethodInvocationExpressionNode`, we would have to check whether it is a standard library call or not. If it is, we can use the `method : StandardLibraryMethod` attribute of `StandardLibraryMethodNode` to determine the correct method.
> 
> If you don't want to make this test with `invocation.getDefinition() instanceof StandardLibraryMethodNode` (or a small visitor), we can add a flag to `MethodNode` or something like that.

OK, I didn't notice that Enum existed, as it isn't part of this PR. I think the test you described should do the job. Additionally, we can add a method to `StandardLibraryEntities` which returns the corresponding `Entity` for each entry of the Enum.

g4rgoil (approved):

PR #67, opened 2021-11-25
Context for method transformation

N-Maas:


src/main/java/edu/kit/compiler/transform/TypeMapper.java
```diff
@@ -96,6 +96,17 @@ public Type getDataType(DataType type) {
         };
     }
 
+    /**
+     * Returns the corresponding Firm `PointerType` for the given class.
+     * 
+     * @param name the class name
+     * @return the corresponding Firm type
+     */
+    public Type getClassPointerType(int name) {
```
g4rgoil, line 105: This should probably be a method of `ClassEntry`, i.e. `ClassEntry::getPointerType()`.

src/main/java/edu/kit/compiler/transform/TransformContext.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+        this.paramTypes = params.toArray(new Type[]{});
+        this.returnType = Optional.empty();
+        if (!methodNode.getType().equals(DataType.voidType())) {
+            this.returnType = Optional.of(typeMapper.getDataType(methodNode.getType()));
+        }
+        Entity methodEntity = typeMapper.getClassEntry(classNode).getMethod(methodNode);
+        Graph graph = new Graph(methodEntity, n_vars);
+        this.construction = new Construction(graph);
+        this.projArgs = construction.newProj(graph.getStart(), Mode.getT(), 2);
```
g4rgoil, line 97: There is a helper method for this in jFirm (`graph.getArgs()`).

src/main/java/edu/kit/compiler/transform/TransformContext.java
```diff
@@ -0,0 +1,175 @@
...
+    private Optional<Type> returnType;
+
+    /**
+     * Current construction object.
+     */
+    @Getter
+    private Construction construction;
+
+    public TransformContext(TypeMapper typeMapper, ClassNode classNode, MethodNode methodNode,
+                            Map<Integer, Integer> variableMapping, int n_vars, boolean isStatic) {
```
g4rgoil, line 73: Is the `n_vars` parameter still necessary? It should always be the same as `variableMapping.size()`.

N-Maas: Well, I wasn't sure whether we want to always use it this way or leave it open to the LocalVariableCounter. However, the counter doesn't return an explicit size either, so there's probably no reason to keep it

g4rgoil (approved):
Looks good to me. Having to manually query the parameter and return types is a bit unfortunate. However, that's a weakness of the `TypeMapper`, which does _not know_ that method entities are of `MethodType`.

PR #69, opened 2021-11-26
Implement standard library and code generation

g4rgoil:
Standard library methods have been tested in small handmade Firm programs.

### DIFF
--- a/run
+++ b/run
@@ -18,6 +18,8 @@ EXE_PATH=$(find "${SCRIPT_PATH}"/target/java-compiler-*-with-dependencies.jar) |
 # ensure that jFirm can find libfirm library
 export LD_LIBRARY_PATH="${SCRIPT_PATH}/target/lib/"
 
+export STD_LIBRARY_PATH="${SCRIPT_PATH}/lib/stdlib.c"
+
 # enable colored logging if stderr is a terminal
 test -t 2 && export COLOR="on"
 
--- a/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
+++ b/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
@@ -1,6 +1,7 @@
 package edu.kit.compiler;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -31,6 +32,9 @@ import edu.kit.compiler.semantic.ErrorHandler;
 import edu.kit.compiler.semantic.NamespaceGatheringVisitor;
 import edu.kit.compiler.semantic.NamespaceMapper;
 import edu.kit.compiler.semantic.SemanticChecks;
+import edu.kit.compiler.transform.Lower;
+import edu.kit.compiler.transform.TypeMapper;
+import firm.Backend;
 import firm.Firm;
 
 public class JavaEasyCompiler {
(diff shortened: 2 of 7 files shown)


run
```diff
@@ -18,6 +18,8 @@ EXE_PATH=$(find "${SCRIPT_PATH}"/target/java-compiler-*-with-dependencies.jar) |
 # ensure that jFirm can find libfirm library
 export LD_LIBRARY_PATH="${SCRIPT_PATH}/target/lib/"
 
+export STD_LIBRARY_PATH="${SCRIPT_PATH}/lib/stdlib.c"
```
N-Maas, line 21: I think this should also be included in the build script, in case unit tests want to use it

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -171,12 +175,30 @@ private static Result compileFirm(String filePath, Logger logger) {
...
+            Backend.createAssembler(assemblyFile, sourceFile);
+
+            var stdLibrary = System.getenv("STD_LIBRARY_PATH");
+            if (stdLibrary == null || !new File(stdLibrary).exists()) {
+                logger.error("standard library implementation not found");
+                return Result.StandardLibraryError;
+            }
+
+            logger.info("compiling program: 'gcc \"%s\" \"%s\"'", assemblyFile, stdLibrary);
+            Runtime.getRuntime().exec(new String[]{ "gcc", assemblyFile, stdLibrary });
```
N-Maas, line 200: Did you test whether this code works?

src/main/java/edu/kit/compiler/transform/StandardLibraryEntities.java
```diff
@@ -0,0 +1,43 @@
...
+import firm.Ident;
+import firm.MethodType;
+import firm.Mode;
+import firm.PrimitiveType;
+import firm.Program;
+import firm.Type;
+import firm.bindings.binding_typerep.ir_visibility;
+import lombok.Getter;
+
+public enum StandardLibraryEntities {
```
N-Maas, line 13: I'm not sure I understand why this is an enum. Do you intend this class to be used statically or do you intend the instance to be passed as an argument to the code generation functions?

Also you probably need to ensure that Firm is initialized before/when the constructor is called.

g4rgoil: This is a simple way of expressing a Singleton in Java. The constructor is not called until the `Instance` entry is first accessed.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
g4rgoil: Yes, I have have successfully compiled and executed code with this. The binary is written to the current working directory as `a.out`. This seems to be what _mjtest_ expects.

src/main/java/edu/kit/compiler/transform/StandardLibraryEntities.java
g4rgoil: Ok, once #67 is merged I will add that.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
N-Maas: :+1:

run
g4rgoil: I have added the environment variable to the build script. It's going to be difficult to test this in unit tests though. We would need to link the assembly generated by jFirm to the library (`gcc` call in the CLI).

N-Maas (approved):

PR #70, opened 2021-11-27
Control flow transformation

N-Maas:
Note: Tests will fail because we need a singleton for firm initialization.

Currently, it contains some dummy code for testing expressions.

### DIFF
--- a/src/main/java/edu/kit/compiler/transform/IRStatementVisitor.java
+++ b/src/main/java/edu/kit/compiler/transform/IRStatementVisitor.java
@@ -1,0 +1,187 @@
+package edu.kit.compiler.transform;
+
+import edu.kit.compiler.data.AstVisitor;
+import edu.kit.compiler.data.DataType;
+import edu.kit.compiler.data.ast_nodes.ExpressionNode;
+import edu.kit.compiler.data.ast_nodes.StatementNode;
+import edu.kit.compiler.data.ast_nodes.StatementNode.BlockStatementNode;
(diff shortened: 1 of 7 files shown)


src/main/java/edu/kit/compiler/transform/IRStatementVisitor.java
```diff
@@ -0,0 +1,187 @@
...
+        Block nextBlock = con.newBlock();
+        nextBlock.addPred(projFalse);
+        nextBlock.mature();
+        con.setCurrentBlock(nextBlock);
+        return false;
+    }
+
+    private Node evalExpression(ExpressionNode expr) {
+        Construction con = context.getConstruction();
+        if (expr instanceof ExpressionNode.IdentifierExpressionNode) {
```
g4rgoil, line 169: Should this really be part of the statement visitor, or do you plan on removing it, once  you have integrated the expression visitor?

N-Maas: This will be removed with the expression visitor. However, until then it is required for the unit tests to work

N-Maas: Well, I could instead let the tests fail, but I'm not sure whether that is better

g4rgoil: In that case I think it's fine as is. Just noticed `instanceof` and wanted to make sure that wasn't permanent.

g4rgoil (approved):

PR #71, opened 2021-11-28
F/firm expressions

0xNB:


src/main/java/edu/kit/compiler/data/AstVisitor.java
```diff
@@ -20,6 +20,7 @@
     default T visit(BlockStatementNode blockStatementNode) { throw new UnsupportedOperationException(); }
     default T visit(LocalVariableDeclarationStatementNode localVariableDeclarationStatementNode) { throw new UnsupportedOperationException(); }
     default T visit(IfStatementNode ifStatementNode) { throw new UnsupportedOperationException(); }
+    default T visit(StatementNode statementNode) { throw new UnsupportedOperationException(); }
```
g4rgoil, line 23: Why did you add this overload? It seems counterproductive in the visitor pattern.

src/main/java/edu/kit/compiler/transform/FirmGraphFactory.java
```diff
@@ -0,0 +1,53 @@
...
+    private StringTable stringTable;
+    private Graph graph;
+
+    public FirmGraphFactory(NamespaceMapper namespaceMapper, StringTable stringTable) {
+        this.namespaceMapper = namespaceMapper;
+        this.stringTable = stringTable;
+    }
+
+    public Graph createGraphFromAst(ProgramNode ast) {
+        TypeMapper entityMapper = new TypeMapper(namespaceMapper, stringTable);
```
g4rgoil, line 29: You should not be constructing a new `TypeMapper` here. It must only be created once, as it changes global state in Firm.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -193,15 +184,14 @@ private static Result compileFirm(String filePath, Logger logger) {
...
      * @param logger the logger
      * @param namespaceMapper empty namespace mapper
      * @return the AST
      * 
-     * @throws CompilerException
-     * @throws IOException
+     * @throws IOException on file system error
      */
-    private static ProgramNode createAttributedAst(
+    public static ProgramNode createAttributedAst(
```
g4rgoil, line 194: We probably shouldn't be using helper methods in the CLI as public API.

build
```diff
@@ -37,7 +40,7 @@ pushd "$SCRIPT_PATH" > /dev/null
 export LD_LIBRARY_PATH="${SCRIPT_PATH}/target/lib/"
 
 # build the executable with Maven
-mvn package "$@"
+LD_LIBRARY_PATH="${SCRIPT_PATH}/target/lib/" mvn package "$@"
```
g4rgoil, line 43: `LD_LIBRARY_PATH` is already exported as environment variable, you don't need to set it here.

pom.xml
```diff
@@ -11,6 +11,10 @@
     <properties>
         <maven.compiler.source>14</maven.compiler.source>
         <maven.compiler.target>14</maven.compiler.target>
+
+        <ld.library.path>./lib</ld.library.path>
```
g4rgoil, line 15: `LD_LIBRARY_PATH` is not a Maven property, it's an environment variable. Why are you setting it here?

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,132 @@
...
+public class IRExpressionVisitor implements AstVisitor<Node> {
+
+    private final TransformContext context;
+
+    public IRExpressionVisitor(TransformContext context) {
+        this.context = context;
+    }
+
+    @Override
+    public Node visit(ExpressionNode.BinaryExpressionNode binaryExpressionNode) {
```
g4rgoil, line 27: You can explicitly import the nested classes. That would probably make your visitor a bit more concise.

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,132 @@
...
+        return n;
+    }
+
+    private Node handleIntegerExpressionNode(ExpressionNode.ValueExpressionNode valueExpressionNode) {
+        if (valueExpressionNode.getValueType() != IntegerLiteral) {
+            throw new IllegalArgumentException("can only evaluate integer expressions");
+        }
+
+        if (valueExpressionNode.getLiteralValue().isEmpty()) {
+            throw new IllegalArgumentException("cannot get empty integer literal");
```
g4rgoil, line 101: All of these checks have already been done during semantic analysis. We can assume that the program (i.e. the AST) during the transformation phase.

build
```diff
@@ -19,11 +19,14 @@ pushd "${SCRIPT_PATH}/libfirm" > /dev/null
...
 
 # build libfirm using the provided makefile
-make -j variant="$FIRM_VARIANT"
+export CC=gcc
+make -j variant="$FIRM_VARIANT" -Wall
 make_exit=$?
 
 popd > /dev/null
 
+mvn clean "$@"
```
g4rgoil, line 28: Why are you calling Maven clean in the build script?

test
```diff
@@ -0,0 +1,47 @@
...
+
+# change the working directory to the project
+# (mvn -f <path> is not as flexible)
+pushd "$SCRIPT_PATH" > /dev/null
+
+# ensure that jFirm can find libfirm library
+export LD_LIBRARY_PATH="${SCRIPT_PATH}/target/lib/"
+
+# build the executable with Maven
+mvn test "$@"
```
g4rgoil, line 41: Doesn't the `build` script already run tests?

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Override
+    public Node visit(ExpressionNode.ValueExpressionNode valueExpressionNode) {
+        // determine type of value expression and value and set the expression
+        // handle accordingly
+
+        // the new node that should be craeted here
+        Node n = null;
+        switch (valueExpressionNode.getValueType()) {
```
g4rgoil, line 76: The assignment to `n` could be expressed very cleanly with with a `switch` expression.

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+    @Override
+    public Node visit(IdentifierExpressionNode identifierExpressionNode) {
+        Construction con = context.getConstruction();
+        if (identifierExpressionNode.getDefinition().getKind() == DefinitionKind.LocalVariable) {
+            Mode mode = getMode(identifierExpressionNode.getResultType());
+            return con.getVariable(context.getVariableIndex(identifierExpressionNode.getIdentifier()), mode);
+        } else if (identifierExpressionNode.getDefinition().getKind() == DefinitionKind.Parameter) {
+            return context.createParamNode(identifierExpressionNode.getIdentifier());
+        } else {
+            throw new IllegalArgumentException();
```
N-Maas, line 153: You need to handle members, too (similar to `FieldAccess`, by calling the `PointerVisitor`)

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+    @Override
+    public Node visit(BinaryExpressionNode binaryExpressionNode) {
+        Node lhs = binaryExpressionNode.getLeftSide().accept(this);
+        Node rhs = binaryExpressionNode.getRightSide().accept(this);
+
+        Mode m = context.getTypeMapper().getDataType(binaryExpressionNode.getLeftSide().getResultType()).getMode();
+
+        switch (binaryExpressionNode.getOperator()) {
+            // handle stuff with our boolean visitor
+            case Assignment, Equal, LessThanOrEqual,
```
N-Maas, line 36: The expression visitor needs to handle non-boolean assignments

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private Node handleNullExpressionNode() {
+        return getConstruction().newAddress(new Entity(Pointer.NULL));
+    }
+
+    @Override
+    public Node visit(NewObjectExpressionNode newObjectExpressionNode) {
+        TypeMapper.ClassEntry classNode = context.getTypeMapper().getClassEntry(newObjectExpressionNode.getTypeName());
+        int size = classNode.getSize();
+        return callCalloc(1, size, classNode.getClassType());
```
N-Maas, line 185: The result type of `calloc` is a pointer to the class, not the class type itself. You can use
`context.getTypeMapper().getDataType(newObjectExpressionNode.getResultType())` instead or wrap it manually with `new PointerType(...)`

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Override
+    public Node visit(NewArrayExpressionNode newArrayExpressionNode) {
+        Type dataType = context.getTypeMapper().getDataType(newArrayExpressionNode.getElementType());
+        Node arrayLength = newArrayExpressionNode.getLength().accept(this);
+        return callCalloc(arrayLength, dataType.getSize(), dataType);
```
N-Maas, line 192: As for new object, the `dataType` is not the correct result type. You need a pointer to an unsized array of `dataType` (this is also provided by `TypeMapper`)

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+        Node objectAddress = methodInvocationExpressionNode.getObject().get().accept(pointerVisitor);
+        // collect arguments in list with this pointer
+        Node[] arguments = new Node[methodInvocationExpressionNode.getArguments().size() + 1];
+        arguments[0] = objectAddress;
+        for (int i = 1; i < methodInvocationExpressionNode.getArguments().size(); i++) {
+            Node n = methodInvocationExpressionNode.getArguments().get(i).accept(this);
+            arguments[i] = n;
+        }
+        int method = methodInvocationExpressionNode.getName();
+        Node methodAddress = getConstruction().newAddress(IRVisitor.getMethodContexts().get(method).getMethodEntity());
```
N-Maas, line 108: You should construct the method entity by requesting the `ClassNode` for the left side of the method invocation from the `TypeMapper` (or `context.getClassNode`, if there is no left side) and calling `getMethod` on it. (You can get the class name by calling `getIdentifier()` on the type of the left side.)

`getMethodContexts` is quite unnecessary

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Override
+    public Node visit(MethodInvocationExpressionNode methodInvocationExpressionNode) {
+        Node mem = getConstruction().getCurrentMem();
+
+        Mode m = context.getTypeMapper().getDataType(methodInvocationExpressionNode.getResultType()).getMode();
+        if (methodInvocationExpressionNode.getObject().isEmpty()) {
+            throw new IllegalArgumentException("object was null on method call");
+        }
+        Node objectAddress = methodInvocationExpressionNode.getObject().get().accept(pointerVisitor);
```
N-Maas, line 99: You sholdn't call the pointer visitor here, this will crash for most cases. Instead, call `accept(this)` (which might call the pointer visitor recursively)

src/main/java/edu/kit/compiler/transform/IRVisitor.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class IRVisitor implements AstVisitor<Void> {
+
+    private final TypeMapper typeMapper;
+    @Getter
+    private static final Map<Integer, TransformContext> methodContexts = new HashMap<>();
```
N-Maas, line 20: Please remove this (see expression visitor comment)

src/main/java/edu/kit/compiler/transform/IRStatementVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+            ExpressionNode.IdentifierExpressionNode id = (ExpressionNode.IdentifierExpressionNode) expr;
+            if (id.getDefinition().getKind() == DefinitionKind.LocalVariable) {
+                Mode mode = getMode(id.getResultType());
+                return con.getVariable(context.getVariableIndex(id.getIdentifier()), mode);
+            } else if (id.getDefinition().getKind() == DefinitionKind.Parameter) {
+                return context.createParamNode(id.getIdentifier());
+            } else {
+                throw new UnsupportedOperationException();
+            }
+        }
```
N-Maas, line 153: Note that the above lines can be removed with the expression visitor available

src/main/java/edu/kit/compiler/transform/IRBooleanExpressions.java
```diff
@@ -0,0 +1,297 @@
...
+        return expr.accept(new ValueOrConditionalDecisionVisitor());
+    }
+
+    private static Node evalExpression(TransformContext context, ExpressionNode expr) {
+        if (expr.getResultType().equals(boolType) &&
+            (expr instanceof BinaryExpressionNode || expr instanceof UnaryExpressionNode)
+        ) {
+            return IRBooleanExpressions.asValue(context, expr);
+        }
+        return context.getConstruction().newConst(1, Mode.getBu());
```
N-Maas, line 293: This can/should be replaced with a call to the expression visitor

src/main/java/edu/kit/compiler/transform/IRPointerVisitor.java
```diff
@@ -0,0 +1,75 @@
...
+        Construction con = context.getConstruction();
+        if (expr instanceof ThisExpressionNode) {
+            return context.createThisNode();
+        } else if (expr instanceof IdentifierExpressionNode) {
+            return context.createParamNode(((IdentifierExpressionNode)expr).getIdentifier());
+        } else if (expr instanceof ValueExpressionNode) {
+            return con.newConst(0, Mode.getIs());
+        } else {
+            return con.newConst(0, context.getThisPtrType().getMode());
+        }
```
N-Maas, line 73: All of this can/should be replaced with a call to the expression visitor

src/main/java/edu/kit/compiler/transform/Lower.java
```diff
@@ -1,12 +1,38 @@
...
+     * @param construction
+     * @return
+     */
+    public static Node getCallocNode(Construction construction) {
+        SegmentType global = Program.getGlobalType();
+        if (calloc == null) {
+            MethodType callocType = new MethodType(new Type[]{ TypeMapper.getIntegerType(), TypeMapper.getIntegerType()},
+                    new Type[]{ TypeMapper.getPointerType()});
+
+            Entity callocEntity = new Entity(global, "calloc", callocType);
```
N-Maas, line 29: You probably should call `Ident.mangleGlobal("calloc")`. (At least we do this for the standard library methods.)

N-Maas (changes requested):

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+        };
+    }
+
+    private Node handleIntegerExpressionNode(ValueExpressionNode valueExpressionNode) {
+        TargetValue tar_val = new TargetValue(valueExpressionNode.getLiteralValue().get().intValue(), Mode.getIs());
+        return getConstruction().newConst(tar_val);
+    }
+
+    private Node handleNullExpressionNode() {
+        return getConstruction().newAddress(new Entity(Pointer.NULL));
```
N-Maas, line 178: This does not work. Use `getConstruction().newConst(0, Mode.getP())` instead

src/main/java/edu/kit/compiler/transform/IRExpressionVisitor.java
```diff
@@ -0,0 +1,222 @@
...
+            }
+            default -> throw new UnsupportedOperationException("not supported " + unaryExpressionNode.getOperator());
+        }
+    }
+
+    @Override
+    public Node visit(MethodInvocationExpressionNode methodInvocationExpressionNode) {
+        Node mem = getConstruction().getCurrentMem();
+
+        Mode m = context.getTypeMapper().getDataType(methodInvocationExpressionNode.getResultType()).getMode();
```
N-Maas, line 95: This doesn't work for `void` methods, because void can't be mapped. You need to handle these separately

N-Maas (commented):
I have included a few test files, but we probably want to relocate them when setting up mjtest for `compile-firm`

N-Maas (approved):

g4rgoil (approved):

PR #78, opened 2021-12-02
Add constant folding optimization

larsk21:
Add optimization that finds constant values for value nodes where possible and replaces these node with constant nodes in the Firm graph.


larsk21:
To do:
- [ ] add tests (requires working graph construction, especially for expressions)
- [ ] implement remaining operations on constant values
- [ ] figure out how to handle control flow (if we can do that with constant folding)

src/main/java/edu/kit/compiler/optimizations/ConstantOptimization.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Transform the given node with the given associated lattice element if
+     * that element is constant.
+     */
+    private void transform(Node node, TargetValueLatticeElement value) {
+        if (value.isConstant()) {
+            Node constantNode = graph.newConst(value.getValue());
+            Graph.exchange(node, constantNode);
```
g4rgoil, line 40: I think your may run into issues here. For example _Div_ nodes are of mode `T` and can't be replaced directly with constants (_Mod_ is the same). I don't know if there is a built-in way of doing this, but you could use back edges to replace the memory and result projections.

larsk21: Yeah, known issue, will fix it.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -184,6 +190,12 @@ private static Result compileFirm(String filePath, Logger logger) {
             ast.accept(irv);
             Lower.lower(irv.getTypeMapper());
 
+            for (Optimization optimization : optimizations) {
+                for (Graph graph : Program.getGraphs()) {
+                    optimization.optimize(graph);
```
N-Maas, line 195: If we swap the order of the two loops and additionally let an optimization return whether anything changed or not, we could do a generic fixpoint iteration that includes all optimizations

N-Maas (commented):
The functionality looks good to me and all tests from the central mjtest repo succeed :+1:

src/main/java/edu/kit/compiler/io/StackWorklist.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Create a new StackWorklist.
+     */
+    public StackWorklist() {
+        stack = new Stack<>();
+        uniqueElements = true;
+    }
+
+    private Stack<T> stack;
```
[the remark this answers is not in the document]
larsk21: Interesting, I didn't know that!

N-Maas (approved):

PR #88, opened 2021-12-06
Fix division overflow

N-Maas:


g4rgoil (approved):
Seems to work, at least from what I've tested. With a combination of constant folding and arithmetic identities, we may still manage to optimize some simple divisions away.

PR #90, opened 2021-12-07
Add arithmetic identities optimization

g4rgoil:
A list of what is being optimized can be found at the top of the main source file.

Lower level identities (e.g. `x * 2^n  -->  x << n`) have been removed. Such replacements may complicate later optimizations, like using `lea` for multiplications. They could instead be done during instruction selection.


src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -0,0 +1,443 @@
...
+            assert node.getOpCode() == ir_opcode.iro_Div
+                    || node.getOpCode() == ir_opcode.iro_Mod;
+
+            for (var edge : BackEdges.getOuts(node)) {
+                if (edge.node.getMode().equals(Mode.getM())) {
+                    exchange(edge.node, mem);
+                } else if (edge.node.getMode().equals(val.getMode())) {
+                    exchange(edge.node, val);
+                } else {
+                    throw new UnsupportedOperationException(
```
larsk21, line 337: Wouldn't it be enough to simply not exchange the node?

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -0,0 +1,443 @@
...
+                // (x * c) * d --> x * (c * d)
+                foldMultiplication(node);
+            }
+        }
+
+        @Override
+        public void visit(Div node) {
+            if (isConstOne(node.getRight())) {
+                // x / 1 --> x
+                exchangeDivOrMod(node, node.getMem(), node.getLeft());
```
larsk21, line 242: Is it possible that `node` and `node.getMem()` are in different blocks? If so, would we get problems when exchanging the memory successor of the node with its memory predecessor?

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
g4rgoil: Yes, however if we were to start using the control flow projections of `Div` and `Mod`, this code might break in a less predictable way. I'm just being a bit defensive here.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
g4rgoil: I don't think so, `exchange` moves all references from the successor `Mem` to the predecessor. I've tested it with the following code and it the optimization works as expected.

```java
public int div(int x, int y) {
    if (x < 10) {
        return x / 1;
    } else {
        return y / 1;
    }
}
```

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -0,0 +1,447 @@
...
+
+        @Override
+        public void visit(Conv node) {
+            // ? make this more generic??
+            if (node.getOp().getOpCode() == ir_opcode.iro_Conv) {
+                // remove casts introduced by 64 bit division where possible
+                var op = (Conv) node.getOp();
+                if (node.getMode().equals(Mode.getIs())
+                        && op.getMode().equals(Mode.getLs())
+                        && op.getOp().getMode().equals(Mode.getIs())) {
```
N-Maas, line 272: Maybe add the reverse case, too? (Ls -> Is -> Ls)

I think this could occur for chained divisions.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -0,0 +1,447 @@
...
+        /**
+         * Returns the first const predecessor of `node`.
+         */
+        private static TargetValue getConstOperand(Node node) {
+            for (var pred : node.getPreds()) {
+                if (pred.getOpCode() == ir_opcode.iro_Const) {
+                    return ((Const) pred).getTarval();
+                }
+            }
+            return null;
```
N-Maas, line 356: As the method is private, this is rather a stylistic preference (feel free to ignore it, if you are of different opinion), but I would prefer either returning an `Optional` or, alternatively, throwing an exception (which would require checking at the call site that this case is impossible).

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -0,0 +1,447 @@
...
+        /**
+         * Returns the first non-const predecessor of `node`.
+         */
+        private static Node getRestNode(Node node) {
+            for (var pred : node.getPreds()) {
+                if (pred.getOpCode() != ir_opcode.iro_Const) {
+                    return pred;
+                }
+            }
+            return null;
```
N-Maas, line 368: See `getConstOperand`

N-Maas (commented):
Looks good apart from these minor points.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
g4rgoil: Removing such a chain of conversions would change the semantics of the program. The value after the two conversions may differ from the original value.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
g4rgoil: I generally agree with you that returning `null` should be avoided. However, in this case of a private helper-method, I believe it's tolerable and allows the code to be a bit more concise.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
N-Maas: Good point

N-Maas (approved):

PR #92, opened 2021-12-13
Fix high level fix point interation for opt.s

larsk21:


larsk21:
It is probably better to go with this approach for now, since the many new Nodes created during the arithmetic identities optimization make it impractical to use during constant folding analysis (where we do not want to change the graph yet).

N-Maas (approved):

PR #94, opened 2021-12-14
Register allocation

N-Maas:


src/main/java/edu/kit/compiler/register_allocation/ApplyAssignment.java
```diff
@@ -0,0 +1,658 @@
...
+     * (used registers) for creating the function prolog and epilog.
+     */
+    public AssignmentResult doApply() {
+        result = new ArrayList<>();
+        LifetimeTracker tracker = new LifetimeTracker();
+        Map<Integer, String> replace = new HashMap<>();
+
+        int i = 0;
+        for (Block b: ir) {
+            output(".L%d:", b.getBlockId());
```
larsk21, line 93: This is specific to the ELF file format. If possible, the register allocation should not depend on the output format. (Maybe we could add a generic label instruction that is transformed by the assembly writer?)

N-Maas: I don't think we need to support other file formats?

I could of course do some special syntax for block labels instead. Note that in this case we would need to do that for jumps, too. However, would a custom syntax even be better then ELF syntax?

larsk21: We don't need to, but the question is how much we want to hardcode our compiler to certain formats. We could of course use the ELF syntax and transform it to another, when we want to add other file formats. Converting between output formats is not the cleanest solution, but it would suffice for now, I think.

The `Lower` class tests for `Platform.isMac() || Platform.isWindows()`, which is why I thought of it.

PR #96, opened 2021-12-16
Instruction Selection

g4rgoil:
Implements a pattern matching based approach for instruction selection.


g4rgoil:
At this point it should be possible to generate instructions for most MiniJava programs. The following is an incomplete list of functionality, which is still WIP.

- [ ] Central map that keeps track of register sizes
- [ ] Clean-up of operand patterns to allow for complex ones
- [ ] More complex patterns, e.g. using immediates directly, instead of loading them first
- [ ] Special handling for standard library calls

src/main/java/edu/kit/compiler/codegen/PatternCollection.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        @Override
+        public InstructionMatch getMatch(Node node) {
+            return subject.getMatch(node);
+        }
+
+        @Override
+        public int getNewRegister() {
+            return super.getNewRegister();
```
N-Maas, line 135: Isnt't this equivalent to not overriding the method at all?

N-Maas (commented):
Looks quite complete (at least for basic patterns). How much testing did you do?

As a general remark, I find it a bit hard to follow the many different classes/interfaces/abstract classes. It would be helpful to at least have a bit documentation what each class represent/is supposed to do (not necessarily for the concrete pattern implementations, but for the more abstract things).

src/main/java/edu/kit/compiler/codegen/MatcherState.java
```diff
@@ -0,0 +1,95 @@
...
+        matches.put(node.getNr(), match);
+    }
+
+    public void walkTopological(MatchVisitor walker) {
+        graph.incVisited();
+        walkTopological(walker, graph.getEnd());
+    }
+
+    private void walkTopological(MatchVisitor walker, Node node) {
+        // based on GraphBase#walkTopologicalHelper of jFirm
```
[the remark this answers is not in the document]
g4rgoil: I'm not walking the firm graph here. During the first phase of instruction selection a graph of matches is built which closely resembles the firm graph, but is not identical (e.g. it may lack nodes like constants for example). In the second phase this tree is visited in topological order. That's why I needed to replicate this function.

src/main/java/edu/kit/compiler/codegen/InstructionSelection.java
```diff
@@ -0,0 +1,143 @@
...
+                    entry.setExitCondition(match.getCondition());
+                }
+                default -> {
+                }
+            }
+        }
+    }
+
+    @RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+    private final class MatchingVisitor extends NodeVisitor.Default {
```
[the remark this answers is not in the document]
g4rgoil: Fair enough, I'll think of some better names.

src/main/java/edu/kit/compiler/codegen/ExitCondition.java
```diff
@@ -0,0 +1,144 @@
...
+        private final Mode mode;
+
+        private BasicBlocks.BlockEntry trueBlock;
+        private BasicBlocks.BlockEntry falseBlock;
+
+        @Override
+        public List<Instruction> getInstructions() {
+            assert trueBlock != null && falseBlock != null;
+
+            // todo what about unsigned, i.e. booleans?
```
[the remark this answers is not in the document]
g4rgoil: That comment seems to be outdated. Relations for unsigned values should be handled correctly.

src/main/java/edu/kit/compiler/codegen/PhiInstruction.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public final class PhiInstruction {
+    
+    private final List<Entry> entries = new ArrayList<>();
+
+    @Getter
+    private final int destination;
```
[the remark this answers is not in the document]
g4rgoil: I can adopt `target` everywhere.

src/main/java/edu/kit/compiler/codegen/PatternCollection.java
g4rgoil: Yes, this is already fixed locally.

src/main/java/edu/kit/compiler/codegen/pattern/InstructionMatch.java
```diff
@@ -0,0 +1,136 @@
...
+        public abstract PhiInstruction getPhiInstruction();
+
+        @Override
+        public void accept(MatchVisitor visitor) {
+            visitor.visit(this);
+        }
+    }
+
+    @RequiredArgsConstructor(access = AccessLevel.PROTECTED)
+    public static abstract class Some extends Match.Some implements InstructionMatch {
```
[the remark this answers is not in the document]
g4rgoil: The various `Some` classes are not really necessary. Their only point is to return `true` in `matches()`. The rest of the inheritance is necessary for the visitor pattern implemented for `InstructionMatch`.

g4rgoil:
> Looks quite complete (at least for basic patterns). How much testing did you do?
> 
> As a general remark, I find it a bit hard to follow the many different classes/interfaces/abstract classes. It would be helpful to at least have a bit documentation what each class represent/is supposed to do (not necessarily for the concrete pattern implementations, but for the more abstract things).

Testing has so far mostly been done manually, i.e. running the code for small examples an checking the output. I'm not really sure how to unit test this stuff properly.

As for the lack of documentation, I'm aware of this issue. At this point the I'm pretty satisfied with the inheritance structure. I'll add some documentation for the important parts soon.

src/main/java/edu/kit/compiler/codegen/InstructionSelection.java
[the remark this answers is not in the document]
g4rgoil: The `MatchVisitor` is now called `InstructionMatchVisitor`, which more closely reflects its purpose. Does that resolve your concerns?

src/main/java/edu/kit/compiler/codegen/pattern/DivisionPattern.java
```diff
@@ -27,10 +28,12 @@ public InstructionMatch match(Node node, MatcherState matcher) {
 
             var lhs = left.match(node.getPred(1), matcher);
             var rhs = right.match(node.getPred(2), matcher);
-            var destination = matcher.getNewRegister();
+
+            // todo is hard coding the register size the best choice here?
+            var targetRegister = matcher.getNewRegister(RegisterSize.QUAD);
```
N-Maas, line 33: This is probably a mistake; unlike the inputs, the target register of a `div` needs to have size `DOUBLE`. The conversion from `QUAD` to `DOUBLE` happens implicitly when accessing the register (see next comment)

src/main/java/edu/kit/compiler/codegen/pattern/ConversionPattern.java
```diff
@@ -0,0 +1,84 @@
...
+        public Stream<Node> getPredecessors() {
+            return match.getPredecessors();
+        }
+
+        public String getCmd() {
+            // So far only Is -> Ls and Ls -> Is is implemented
+            if (from.equals(Mode.getIs()) && to.equals(Mode.getLs())) {
+                return "movsl";
+            } else if (from.equals(Mode.getLs()) && to.equals(Mode.getIs())) {
+                return "mov";
```
N-Maas, line 78: This won't work: because the size of the input register is `QUAD`, the register allocation would generate something like `movl %rax, %eax`, which is illegal. If you want to explicitely generate downward-conversions, we would either need a special case in `Instruction` or you need to use something like `lea`.

However, as far as I know this currently applies only to `div` instructions, where the register allocation already creates the appropriate move for the target register implicitly. So you should be fine by just not creating an instruction at all in this case

src/main/java/edu/kit/compiler/codegen/InstructionSelection.java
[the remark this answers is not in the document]
N-Maas: Yes, seems ok to me

src/main/java/edu/kit/compiler/codegen/pattern/ConversionPattern.java
g4rgoil: You're right, this is currently generating illegal code. I think your suggestion of making this a noop and changing the size of division target registers to `DOUBLE` should work.
Note that this will generate a _useless_ temporary register due to the `Conv` node after `Div`. Would that be an issue for the register allocation?

N-Maas: No, any virtual registers that aren't used will be ignored by the register allocation. This works even for the trivial allocator

g4rgoil: So, the original solution would not have worked with a noop. The `Conv` node now reuses the result register of the division. The fundamental issue seems to be that our virtual register are assigned a static size and can therefore not be used in different contexts.
If we want to fix/improve this, we could specify the desired size, every time the register is used. Alternatively, I could imagine adding a special instruction type for conversions.

N-Maas: I would definitively prefer keeping static register sizes

src/main/java/edu/kit/compiler/codegen/pattern/InstructionMatch.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.stream.Stream;
+
+import edu.kit.compiler.codegen.ExitCondition;
+import edu.kit.compiler.codegen.PhiInstruction;
+import edu.kit.compiler.intermediate_lang.Instruction;
+import firm.nodes.Node;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
+
+public interface InstructionMatch extends Match {
```
N-Maas, line 15: As this is an interface now, you can remove the `public`, `abstract` and `static`modifiers

N-Maas:
Generally looks good to me, except that some documentation for the most important classes/interfaces would be good

src/main/java/edu/kit/compiler/codegen/pattern/InstructionMatch.java
g4rgoil: I definitely can't remove the `static` modifiers, but you're right about the other two.

N-Maas (approved):

PR #98, opened 2021-12-21
Better patterns for Instruction Selection

g4rgoil:
Extends instruction selection with a number of more complex patterns. Currently includes the at least the following:

(Note: _Ill-fated attempts at arithmetic replacements have been remove and will be part of a separate PR._)

- Verify size of immediate values (e.g. 64-bit only allowed in move to register)
- Inline immediate values where possible
- Reorder operands where allowed to be able to inline more immediates
- Use x86's addressing modes where possible. Supports combinations of static offset, base register, index register, and index scaling factors
- Extend operands to load immediate values only once needed. Previously, every immediate used at any point of a function was loaded as part of its Start block
- Handle comparisons to zero specially, using the `test` instead of `cmp` instruction


src/main/java/edu/kit/compiler/codegen/pattern/OperandPattern.java
```diff
@@ -66,34 +121,208 @@
...
+            var indexLeft = matchMul(nodes.offset, nodes.secondRegister,
+                    nodes.firstRegister, matcher);
+            if (indexLeft.matches()) {
+                return indexLeft;
+            }
+
+            // todo (2,4,8) + 1 * %rax = (%rax,%rax,8)
+            var match = getMatch(nodes.offset, nodes.firstRegister,
+                    nodes.secondRegister, matcher);
+            if (match == null) {
```
N-Maas, line 198: I might be mistaken, but to me it doesn't seem possible that `getMatch` returns `null`. Maybe this should rather be `match.matches()`?

src/main/java/edu/kit/compiler/codegen/pattern/OperandPattern.java
```diff
@@ -66,34 +121,208 @@
...
+                return indexRight;
+            }
+
+            var indexLeft = matchMul(nodes.offset, nodes.secondRegister,
+                    nodes.firstRegister, matcher);
+            if (indexLeft.matches()) {
+                return indexLeft;
+            }
+
+            // todo (2,4,8) + 1 * %rax = (%rax,%rax,8)
```
N-Maas, line 195: Is this todo up to date?

src/main/java/edu/kit/compiler/codegen/pattern/OperandPattern.java
```diff
@@ -66,34 +121,208 @@
...
+                if (scaleMatch.matches()) {
+                    return getMatch(Optional.ofNullable(offset),
+                            Optional.ofNullable(base), Optional.of(mul.getLeft()),
+                            Optional.of(scaleMatch.getOperand()), matcher);
+                } else {
+                    return OperandMatch.none();
+                }
+            } else {
+                return OperandMatch.none();
+            }
```
N-Maas, line 229: You could eliminate one branch by using an unconditional `return` at the end of the method.

src/main/java/edu/kit/compiler/codegen/pattern/BinaryInstructionPattern.java
```diff
@@ -16,30 +16,43 @@
...
+    public static <T extends Operand.Target, S extends Operand.Source> BinaryInstructionPattern<T, S> of(
+            ir_opcode opcode, String command, Pattern<? extends OperandMatch<T>> target,
+            Pattern<? extends OperandMatch<S>> source, int offset, boolean commutate) {
+        return new BinaryInstructionPattern<T, S>(opcode, target, source, offset, commutate) {
+            @Override
+            protected InstructionMatch getMatch(Node node, OperandMatch<T> target,
+                    OperandMatch<S> source, MatcherState matcher) {
+                var targetRegister = getTarget(target.getOperand(),
+                        () -> matcher.getNewRegister(source.getOperand().getSize()));
+                return new BinaryInstructionMatch(node, command, target, source, targetRegister);
```
N-Maas, line 37: Is there a specific reason why this is an anonymous subclass instead of `command` being a member of the class?

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -346,13 +345,15 @@ private static void createOptimizedIR(String filePath, Logger logger,
         Lower.lower(irv.getTypeMapper());
 
         for (Graph graph : Program.getGraphs()) {
+            Dump.dumpGraph(graph, "before");
```
N-Maas, line 348: This is probably a mistake

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -346,13 +345,15 @@ private static void createOptimizedIR(String filePath, Logger logger,
...
             boolean changed;
             do {
                 changed = false;
                 for (Optimization optimization : optimizations) {
                     changed |= optimization.optimize(graph);
                 }
             } while (changed);
+            Dump.dumpGraph(graph, "after");
```
N-Maas, line 356: --

N-Maas (commented):
Seems to work fine and specifically the improved immediate handling should be really helpful for performance

src/main/java/edu/kit/compiler/codegen/pattern/OperandPattern.java
g4rgoil: Yes, it's a further optimization that could be implemented in the future. It would only be useful in combination with support for `lea` though.

src/main/java/edu/kit/compiler/codegen/pattern/OperandPattern.java
g4rgoil: I'm aware of that, but I try to avoid using fallthrough wherever possible, as it can make code more complex to read (obviously in this case it's pretty unnecessary).

src/main/java/edu/kit/compiler/codegen/pattern/BinaryInstructionPattern.java
g4rgoil: This is a remnant of the patterns I tried to implement for Mul and Div replacement. Seems I forgot to remove this change to `BinaryInstruction` with the rest of the changes.

N-Maas (approved):

PR #99, opened 2021-12-22
F/backend/phi

0xNB:


src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
```diff
@@ -71,14 +72,14 @@ public int getLabel() {
          * Return the basic instructions of this block.
          */
         public List<Instruction> getInstructions() {
-            return Collections.unmodifiableList(instructions);
+            return instructions;
```
g4rgoil, line 75: There is a reason why this originally returned an unmodifiable list. You're not really supposed to change the state of this class. For example, I'm currently considering breaking `instructions` into two lists. You're change here makes that pretty much impossbile.

src/main/java/edu/kit/compiler/codegen/PhiRemover.java
```diff
@@ -0,0 +1,72 @@
...
+    public PhiRemover(BasicBlocks basicBlocks) {
+        this.basicBlocks = basicBlocks;
+    }
+
+    /**
+     * transforms this list of blocks into a new one without any phi instructions
+     *
+     * @return list of basic blocks without phis
+     */
+    public BasicBlocks removePhis() {
```
g4rgoil, line 35: I would have expected your code to receive an instance of `BasicBlocks` and return some kind of `Map<Integer, List<Instruction>>` (which is what we need for register allocation).

src/main/java/edu/kit/compiler/codegen/PhiRemover.java
```diff
@@ -0,0 +1,72 @@
...
+                    PhiSourceNodeRegister targetRegister = new PhiSourceNodeRegister(phiInstruction.getTargetRegister(), nullNode);
+                    permutationSolver.addMapping(sourceRegister, targetRegister);
+                    int max = Math.max(source.getRegister(), phiInstruction.getTargetRegister());
+                    if (max > maxRegister) {
+                        maxRegister = max + 1;
+                    }
+                }
+            }
+            // solve permutations
+            List<PhiAssignment> permutations = permutationSolver.solve(new PhiSourceNodeRegister(maxRegister, nullNode));
```
g4rgoil, line 53: You seem to be using `maxRegister` for spilling. However, as far as I can tell, there is no guarantee, that register isn't used elsewhere. I'm sure I can expose my register counter to you somehow, so you can properly allocate the temporary register.

g4rgoil (changes requested):
Are currently dealing with the instruction generated by `ExitCondition` in some way?

N-Maas:
Replaced by #103

PR #100, opened 2021-12-22
Add assembly file writer

larsk21:


src/main/java/edu/kit/compiler/assembly/ElfAssemblyWriter.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private void writeFunction(FunctionInstructions function) {
+        writeFunctionProlog(function.getLdName());
+
+        for (String line : function.getInstructions()) {
+            String[] commandComment = split(line);
+            String command = commandComment[0];
+            String comment = commandComment[1];
+
+            if (command.equals(ApplyAssignment.FINAL_BLOCK_LABEL)) {
```
N-Maas, line 104: `FINAL_BLOCK_LABEL` is only a substring of the command. Currently, the register allocator outputs `FINAL_BLOCK_LABEL + ":"` for the label itself and `"jmp " + FINAL_BLOCK_LABEL` for jumps

N-Maas (approved):

PR #101, opened 2021-12-29
Ensure that replacing virtual registers works correctly

N-Maas:
Also adds an additional check for instruction creation and fixes a missing input register in a memory pattern


g4rgoil (approved):
Good job catching the missing input registers.

PR #103, opened 2021-12-29
Phi resolver and block ordering

N-Maas:
Currently, the phi resolver interacts with the instruction selection in the following ways:

1. using the state of the instruction selection to create new virtual registers
2. using the state of the instruction selection to create new block ids
3. appending move instructions to some blocks
4. altering exit conditions when a new block is introduced

1 and 2 are probably unavoidable, 3 could also be done within the `PhiResolver` instead. However, while 4 could be done separately, it would be rather ugly to do so, as the `PhiResolver` would need to duplicate the exit condition mechanism of the instruction selection somehow.

Also, the block ids are now serialized.

### DIFF
--- a/src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
+++ b/src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
@@ -52,6 +52,10 @@ public class BasicBlocks {
         return blocks.values();
     }
 
+    public BlockEntry getStartBlock() {
+        return blocks.get(graph.getStartBlock().getNr());
+    }
+
     @Override
     public String toString() {
         return graph.getEntity().getName() + ": " + blocks.toString();
--- a/src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
+++ b/src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
@@ -118,9 +118,9 @@ public class BasicBlocks {
             var condition = exitCondition.get();
             if (node.getPred().getOpCode() == ir_opcode.iro_Cond) {
                 if (node.getNum() == Cond.pnTrue) {
-                    condition.setTrueBlock(entry);
+                    condition.setTrueBlock(entry.getBlockId());
                 } else if (node.getNum() == Cond.pnFalse) {
-                    condition.setFalseBlock(entry);
+                    condition.setFalseBlock(entry.getBlockId());
                 } else {
                     throw new IllegalStateException();
                 }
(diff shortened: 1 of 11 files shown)


src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
```diff
@@ -54,17 +73,20 @@ public String toString() {
 
         @Getter
         private final Block firmBlock;
+        @Getter
+        private final int blockId;
```
g4rgoil, line 77: Calling this `blockId` might be a bit misleading, as it is not the value used to access the entry in `BasicBlocks`. I think something like `jumpLabel` might be a better fit. Please correct me if I have misunderstood what the field is used for.

src/main/java/edu/kit/compiler/codegen/ExitCondition.java
```diff
@@ -76,38 +92,53 @@ public void setFalseBlock(BlockEntry block) {
         private final Operand.Source target;
         private final Operand.Source source;
 
-        private BasicBlocks.BlockEntry trueBlock;
-        private BasicBlocks.BlockEntry falseBlock;
+        private int trueBlockId = -1;
+        private int falseBlockId = -1;
```
g4rgoil, line 96: Why the switch to using integers here? I know using `null` for missing values is not great, I'm not sure using `-1` instead is much of an improvement.

src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
N-Maas: Fair

src/main/java/edu/kit/compiler/codegen/ExitCondition.java
N-Maas: The intention behind this change is not about `null`, instead the reason is that not every block that could be referenced here has a `BlockEntry`. Specifically, new blocks that are created for critical edges by the `PhiResolver` don't get a new entry (but still have an integer label)

src/main/java/edu/kit/compiler/codegen/BasicBlocks.java
g4rgoil: The `getLabel` and `getJumpLabel` methods are somewhat redundant now.

src/main/java/edu/kit/compiler/codegen/ExitCondition.java
g4rgoil: OK, I am can live with this change. The only thing we really lose, is the ability to navigate from a `BlockEntry` to its control flow successors. Though to be fair, we weren't using that to begin with, so it shouldn't be an issue.

PR #104, opened 2022-01-02
Complete the backend

N-Maas:
Implements --compile and adds a selection of the tests from the mjtest repo to the CI


run
```diff
@@ -24,5 +24,6 @@ export STD_LIBRARY_PATH="${SCRIPT_PATH}/lib/stdlib.c"
 test -t 2 && export COLOR="on"
 
 # execute the compiler with the provided arguments and return the exit code
-java -jar "$EXE_PATH" "$@"
+# -ea enables assertions
+java -jar -ea "$EXE_PATH" "$@"
```
g4rgoil, line 28: Should we really enable assertions here (this script is used for benchmarks)?

N-Maas: Assertions should be enabled at least for the test script, I think, but disabled for the final version of the compiler. Regarding the run script here, we could either separate it from the test script and/or remove the assertions later, once we are confident enough in the correctness.

I wonder whether enabled assertions might simplify pinning down bugs found by our supervisors?

g4rgoil: I agree that we should definitely enable assertions in the `test` script (which doesn't currently exist in master). For now we can also leave them enabled in `run`. I'm not sure that should be permanent though. Even ignoring the performance penalty, it risks test failures because of incorrect assertions.

g4rgoil (approved):
Currently the backend passes every test case. I would suggest merging this PR soon, so we can use it as a basis for improvements to instruction selection and register allocation.

PR #106, opened 2022-01-03
Fix incorrect register size in call

N-Maas:


g4rgoil (approved):
Fixes the original issue, also works with `boolean` and Pointer.

PR #107, opened 2022-01-03
F/opt/common subexpressions

0xNB:
common subexpression optimization


src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+
+        @Override
+        public void visit(Conv node) {
+            visitUnary(node, node.getOp(), binding_irnode.ir_opcode.iro_Conv);
+        }
+
+        @Override
+        public void visit(Div node) {
```
g4rgoil, line 266: Do you consider the memory input of `Div` and `Mod` in any way? The language report does state that behavior of division by zero is undefined, but we can't just ignore the resulting exceptions either (at least so far we haven't).

src/test/java/edu/kit/compiler/cmd/EchoLineEndingTest.java
```diff
@@ -22,15 +23,24 @@
             "\n" +
             "lkjasdfkjalsdf";
 
+    // ignore this test on windows
     @Test
+    @Disabled
```
g4rgoil, line 28: This is not really related to this PR. Also you can just delete the entire file, as it is not doing anything anymore.

configure_env
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
g4rgoil, line 1: Do we need this script (same goes for `test`)?

0xNB:
Replace Unary and Binary Map with Map of Variable Length Keys for function calls etc.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,399 @@
...
+            // do nothing
+        }
+
+        @Override
+        public void visit(Call call) {
+            visitPreds(call, iterableToArray(call.getPreds()), binding_irnode.ir_opcode.iro_Call);
+        }
+
+        @Override
+        public void visit(Const node) {
```
g4rgoil, line 309: If you want to eliminate function calls, you also need to handle `Address` nodes in some way. They should be pretty much like `Const`, just using the referenced `Entitiy` as key, instead of the `TargetValue`.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,399 @@
...
+    private final class CSEVisitor extends NodeVisitor.Default {
+
+        private final Map<TargetValue, Node> constCache = new HashMap<>();
+        private final Map<NodePreds, Node> nodeCache = new HashMap<>();
+
+        @RequiredArgsConstructor
+        private class NodePreds {
+            private final Node[] preds;
+            private final binding_irnode.ir_opcode opcode;
+            private final Mode mode;
```
g4rgoil, line 216: The `Mode` is only included in your `equals`. This breaks the contract concerning the behavior of `equals` and `hashCode`. Maybe consider just letting `lombok` generate the two methods for you.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,388 @@
...
+        public void defaultVisit(Node node) {
+            // change nothing
+        }
+
+        private void visitPreds(Node node, Node[] preds, binding_irnode.ir_opcode opcode) {
+            NodePreds np = new NodePreds(preds, opcode, node.getMode());
+            if (nodeCache.containsKey(np)) {
+                Node n = nodeCache.get(np);
+                // can exchange node with already found in list
+                nodeValues.put(node, n);
```
N-Maas, line 241: It might be a good idea to check which of `node` / `n` dominates the other (if any) and put the dominating one in the cache.

N-Maas (commented):
Generally seems to work at least for nodes in the same block. However, you definitively need to clean up the code. There are print statements and graph dumps at multiple locations.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,388 @@
...
+        for (BackEdges.Edge edge : BackEdges.getOuts(n)) {
+            Node node = edge.node;
+            if (node.getOpCode() == binding_irnode.ir_opcode.iro_Call) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private int dominates(Pointer source, Pointer target) {
```
N-Maas, line 174: This should rather return a boolean.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,388 @@
...
+        if (newNode != null) {
+            Pointer newPointer = newNode.getBlock().ptr;
+            Pointer origPointer = orig.getBlock().ptr;
+            Pointer replacementPointer = replacement.getBlock().ptr;
+
+            int dominatesOld = dominates(newPointer, origPointer);
+            int dominatesNew = dominates(newPointer, replacementPointer);
+
+            if (dominatesOld == 0 || dominatesNew == 0) {
+                // can not replace node in this case since new node does not dominate one of the replacements
```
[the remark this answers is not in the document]
0xNB: Both replacements are checked since both nodes are inserted to be checked for similarity here.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,388 @@
...
+            }
+            // new node has to dominate all uses
+            boolean usesDominated = dominatesUses(newNode, orig);
+            boolean usesDominatedRep = dominatesUses(newNode, replacement);
+
+            if (hasCallAsChild(orig) || hasCallAsChild(replacement)) {
+                System.out.println("call as child do not replace");
+                return false;
+            }
+            if (usesDominated && usesDominatedRep) {
```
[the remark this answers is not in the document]
0xNB: `dominates()` checks for block dominance while the use can be in another block that lies before the block of the to be replaced node. This is the case for loops ie.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,388 @@
...
+
+            if (dominatesOld == 0 || dominatesNew == 0) {
+                // can not replace node in this case since new node does not dominate one of the replacements
+                return false;
+            }
+            // new node has to dominate all uses
+            boolean usesDominated = dominatesUses(newNode, orig);
+            boolean usesDominatedRep = dominatesUses(newNode, replacement);
+
+            if (hasCallAsChild(orig) || hasCallAsChild(replacement)) {
```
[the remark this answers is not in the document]
0xNB: This has been indeed fixed and can be removed here.

src/main/java/edu/kit/compiler/optimizations/NodeCollector.java
```diff
@@ -0,0 +1,20 @@
...
+
+import firm.nodes.Node;
+import firm.nodes.NodeVisitor;
+import lombok.Getter;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Getter
+public class NodeCollector extends NodeVisitor.Default {
```
N-Maas, line 11: This class is currently not used. However, I would suggest to keep it and instead remove `WorklistFiller` and `DefaultWorklist`; the CSE doesn't seem to need a proper worklist anyway (it only iterates over the nodes in the list and doesn't reinsert any nodes).

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+            if (!dominatesOld || !dominatesNew) {
+                // can not replace node if the new node isnt dominating both former nodes.
+                return false;
+            }
+            // new node has to dominate all uses
+            boolean usesDominated = dominatesUses(newNode, orig);
+            boolean usesDominatedRep = dominatesUses(newNode, replacement);
+
+            if (usesDominated && usesDominatedRep) {
+                //      System.out.println("replaced node " + orig + " with " + newNode + " in graph " + g);
```
N-Maas, line 149: can be removed

N-Maas (commented):
Generally looks good now

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+    /**
+     * Transform the given node with the given associated lattice element if
+     * that element is constant.
+     */
+    private boolean transform(Graph g, Node orig, Node replacement, binding_irnode.ir_opcode opcode) {
+        Node newNode = switch (opcode) {
+            case iro_Add -> {
+                Add add = (Add) orig;
+                yield g.newAdd(orig.getBlock(), add.getLeft(), add.getRight());
+            }
```
g4rgoil, line 87: You have added a number of new nodes the `CSEVisitor`, don't you need to propagate these changes to `transform`?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+
+        @Override
+        public void visit(Cond node) {
+            // do nothing
+        }
+
+        @Override
+        public void visit(Proj proj) {
+            // do nothing
+        }
```
g4rgoil, line 278: I think you need to handle `Proj`s as well. For example, if you eliminate a pair of `Load`s, you'd also want to unify their result projections. Obviously beware that `Proj`s are only equal if they extract the same number.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    private boolean dominates(Pointer source, Pointer target) {
+        int dominates = 0;
+
+        if (!dominateMap.containsKey(source)) {
+            // we have no entry yet, compute both values
+            dominates = binding_irdom.block_dominates(source, target);
+
+            HashMap<Pointer, Integer> cachedResults = new HashMap<>();
```
g4rgoil, line 177: Are you caching for performance reasons only? If so, note that the implementation of `block_dominates` just dereferences a few pointer, and then compares two integers.

src/main/java/edu/kit/compiler/optimizations/NodeCollector.java
larsk21: There are a number of visitors for filling data structures in `optimizations.Util`. We should probably use those instead.

src/main/java/edu/kit/compiler/io/DefaultWorklist.java
```diff
@@ -0,0 +1,34 @@
+package edu.kit.compiler.io;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DefaultWorklist<T> implements Worklist<T>{
```
larsk21, line 6: What is the benefit of this class?

src/main/java/edu/kit/compiler/optimizations/WorklistFiller.java
```diff
@@ -0,0 +1,26 @@
+package edu.kit.compiler.optimizations;
+
+import edu.kit.compiler.io.Worklist;
+import firm.nodes.Node;
+import firm.nodes.NodeVisitor;
+
+/**
+ * Firm node visitor that inserts all visited nodes in the given Worklist.
+ */
+public class WorklistFiller extends NodeVisitor.Default {
```
larsk21, line 10: Same as for `NodeCollector`, there are visitors for filling data structures in `optimizations.Util`.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+                // for each replacement merge nodes together
+                Node orig = replacement.getKey();
+                Node n = replacement.getValue();
+
+                changes |= transform(g, orig, n, orig.getOpCode());
+            }
+            if (changes) {
+                hadChange = true;
+            }
+            maxChanges++;
```
larsk21, line 65: Should this be inside `if (changes)`?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+            }
+            case iro_Minus -> {
+                Minus minus = (Minus) orig;
+                yield g.newMinus(orig.getBlock(), minus.getOp());
+            }
+            case iro_Const -> {
+                Const cons = (Const) orig;
+                yield g.newConst(cons.getTarval().asInt(), cons.getMode());
+            }
+            case iro_Sel -> {
```
larsk21, line 116: I think out optimizations runs after `lower`, so we shouldn't have `Sel` nodes at this point.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
larsk21: This is correct, Firm caches the dominance information itself.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+                    // replace and update cache
+                    nodeValues.put(n, node);
+                    nodeCache.put(np, node);
+                }
+            } else {
+                nodeCache.put(np, node);
+            }
+        }
+
+        private Node[] iterableToArray(Iterable<Node> iterable) {
```
larsk21, line 247: Please add this as `toArray` to `io.CommonUtil` as a generic function similar to `CommonUtil.toList` (I would suggest using a Stream).

src/test/java/edu/kit/compiler/optimizations/CommonSubexpressionTest.java
```diff
@@ -0,0 +1,129 @@
...
+    private String surroundWithIO(String program, String expression) {
+        return "System.in.read()\n"
+                + program + "\n"
+                + "System.out.write(" + expression + ")";
+    }
+
+
+    private List<Node> getNodes(Graph graph) {
+        List<Node> nodes = new ArrayList<>();
+        NodeVisitor nodeVisitor = new NodeVisitor.Default() {
```
larsk21, line 38: You can use the generic node list filler here.

src/test/java/edu/kit/compiler/optimizations/FirmGraphGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                }
+            }
+        }
+    }
+
+    public static void dumpGraph(Graph g) throws IOException {
+        JFirmSingleton.initializeFirmLinux();
```
larsk21, line 64: You have to assure that this is not called more than once.

src/test/java/edu/kit/compiler/optimizations/FirmGraphGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+                Files.move(currentFile.toPath(), targetFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+
+                currentFile.delete();
+                System.out.printf("moved file %s\n", s);
+            }
+        }
+    }
+
+    public static List<Graph> getFirmGraphs(String inputFile) throws IOException {
+        JFirmSingleton.initializeFirmLinux();
```
larsk21, line 88: see above

src/test/java/edu/kit/compiler/optimizations/FirmGraphGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            ast.accept(irv);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        return StreamSupport.stream(Program.getGraphs().spliterator(), false).collect(Collectors.toList());
+    }
+
+    public static int getGraphCount() {
+        JFirmSingleton.initializeFirmLinux();
```
larsk21, line 125: see above

src/test/java/edu/kit/compiler/optimizations/FirmGraphGenerator.java
N-Maas: Actually, this is already the job of the singleton.

larsk21: Ah, I see.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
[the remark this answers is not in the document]
0xNB: I think this conversation can also be closed? @N-Maas Or if you have another idea we can use that aswell

src/main/java/edu/kit/compiler/optimizations/NodeCollector.java
0xNB: `DefaultWorklist` and `WorklistFiller` have been replaced with the `NodeCollector` instead

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+                Mul mul = (Mul) orig;
+                yield g.newMul(orig.getBlock(), mul.getLeft(), mul.getRight());
+            }
+            case iro_Minus -> {
+                Minus minus = (Minus) orig;
+                yield g.newMinus(orig.getBlock(), minus.getOp());
+            }
+            case iro_Const -> {
+                Const cons = (Const) orig;
+                yield g.newConst(cons.getTarval().asInt(), cons.getMode());
```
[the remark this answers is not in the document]
0xNB: Done

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,387 @@
...
+
+    private boolean hasCallAsChild(Node n) {
+        for (BackEdges.Edge edge : BackEdges.getOuts(n)) {
+            Node node = edge.node;
+            if (node.getOpCode() == binding_irnode.ir_opcode.iro_Call) {
+                return true;
+            }
+        }
+        return false;
+    }
```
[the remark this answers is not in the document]
0xNB: Done

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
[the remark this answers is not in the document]
N-Maas: I'm still not sure whether the check is actually necessary, but in doubt the more conservative option is alright

src/main/java/edu/kit/compiler/io/DefaultWorklist.java
0xNB: Removed

src/main/java/edu/kit/compiler/optimizations/WorklistFiller.java
0xNB: Removed

src/main/java/edu/kit/compiler/optimizations/NodeCollector.java
larsk21: Why create the new class `NodeCollector`, if we already have `Util.NodeListFiller`?

src/test/java/edu/kit/compiler/optimizations/CommonSubexpressionTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // return a + d;
+        // ->
+        // int tmp = b * c;
+        // int a = tmp + g;
+        // int d = tmp * e;
+        // return a + d;
+        String file = "edu/kit/compiler/optimizations/Subexpressions.java";
+        Graph g = build(file);
+
+        Dump.dumpGraph(g, "before-opt");
```
larsk21, line 75: Please remove all `dumpGraph`s before merging

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+        }
+
+        @Override
+        public void visit(Minus node) {
+            visitPreds(node, Util.iterableToArray(node.getPreds()), binding_irnode.ir_opcode.iro_Minus);
+        }
+
+        @Override
+        public void visit(Mod node) {
+            //   visitPreds(node, Util.iterableToArray(node.getPreds()), binding_irnode.ir_opcode.iro_Mod);
```
g4rgoil, line 413: `Mod` and `Div` are still commented out.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+        g.assureProperties(IR_GRAPH_PROPERTY_CONSISTENT_DOMINANCE);
+
+        return hadChange;
+    }
+
+    /**
+     * Transform the given node with the given associated lattice element if
+     * that element is constant.
+     */
+    private boolean transform(Graph g, Node orig, Node replacement, binding_irnode.ir_opcode opcode, HashMap<Node, Node> replacementMap) {
```
g4rgoil, line 108: `Div` and `Mod` are also missing from `transform`.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+                if (dominates(n.getBlock().ptr, node.getBlock().ptr)) {
+                    nodeValues.put(node, n);
+                } else {
+                    nodeValues.put(n, node);
+                    cmpCache.put(cmpPreds, node);
+                }
+            } else {
+                cmpCache.put(cmpPreds, node);
+            }
+        }
```
g4rgoil, line 338: You're not actually doing anything with this information. Also, I don't see any value in eliminating `Cmp`s.

larsk21: In addition, `CmpPreds` does not define an `equals` method, so `cmpCache.containsKey(cmdPreds)` will always be `false`.

0xNB: The nodeValues is updated which exchanges `Cmp` nodes, I have added equals and hashCode here though.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
0xNB: Done

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+                Graph.exchange(addr, newAddr);
+            } else {
+                addressCache.put(ent, addr);
+            }
+        }
+
+        @Override
+        public void visit(Call call) {
+            Attributes a = s.getAttributeAnalysis().getAttributes(Util.getCallee(call));
+            if (a.isPure()) {
```
g4rgoil, line 375: This check is not sufficient. You either need to check `terminates` (or preferably `Util#isPinned`) and `canDeduplicate`.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+            binding_irgopt.remove_bads(g.ptr);
+        } while (changes && maxChanges < 10);
+
+        if (!backEdgesEnabled) {
+            BackEdges.disable(g);
+        }
+        binding_irgopt.remove_bads(g.ptr);
+        binding_irgopt.remove_unreachable_code(g.ptr);
+        binding_irgopt.remove_bads(g.ptr);
+        g.assureProperties(IR_GRAPH_PROPERTY_CONSISTENT_DOMINANCE);
```
larsk21, line 99: Shouldn't this be `confirmProperties(NONE)`?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+            }
+            if (changes) {
+                hadChange = true;
+                maxChanges++;
+            }
+
+            binding_irgopt.remove_bads(g.ptr);
+            binding_irgopt.remove_unreachable_code(g.ptr);
+            binding_irgopt.remove_bads(g.ptr);
+        } while (changes && maxChanges < 10);
```
larsk21, line 91: Is it necessary to have a fix point iteration here where a new `CSEVisitor` is created in every iteration?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+
+            HashMap<Node, Node> replacementMap = new HashMap<>();
+            for (var replacement : nodeValues.entrySet()) {
+                // for each replacement merge nodes together
+                Node orig = replacement.getKey();
+                Node n = replacement.getValue();
+
+                int maxIndirections = 150;
+                while (orig.getOpCode() == binding_irnode.ir_opcode.iro_Deleted && maxIndirections > 0) {
+                    orig = replacementMap.getOrDefault(orig, orig);
```
larsk21, line 65: Given `orig.getOpCode() == iro_Deleted` is `true` and `replacementMap.get(orig)` is empty, then `replacementMap.getOrDefault(orig, orig)` will return `orig` and the condition `orig.getOpCode() == iro_Deleted` will continue to be `true`, right?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,466 @@
...
+
+            CSEVisitor visitor = new CSEVisitor(state);
+
+            List<Node> nodes = new ArrayList<>();
+            Util.NodeListFiller nodeCollector = new Util.NodeListFiller(nodes);
+            g.walkTopological(nodeCollector);
+            while (!nodes.isEmpty()) {
+                Node n = nodes.remove(0);
+                n.accept(visitor);
+            }
```
larsk21, line 55: Is there a difference to `g.walkTopological(visitor)`?

g4rgoil: There does seem to be a difference. Probably because `Address` and `Const` nodes are replaces in the visitor. That should probably moved to `transform`.

larsk21: Good point, at first look I thought the Visitor was only for analysis

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
g4rgoil: The default case should never occur (and in practice doesn't seem to), so I would leave this as it is for now. The `maxIndirection` is actually hit a few times.

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,412 @@
...
+            binding_irgopt.remove_bads(g.ptr);
+        } while (changes && numChanges < MAX_CHANGES);
+
+        if (!backEdgesEnabled) {
+            BackEdges.disable(g);
+        }
+
+        binding_irgopt.remove_bads(g.ptr);
+        binding_irgopt.remove_unreachable_code(g.ptr);
+        binding_irgopt.remove_bads(g.ptr);
```
larsk21, line 87: Is this not done before we leave the loop?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
```diff
@@ -0,0 +1,412 @@
...
+            if (!dominates(newNode.getBlock().ptr, n.getBlock().ptr)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static Node findReplacement(Node node, Map<Node, Node> replacements) {
+        for (int i = 0; i < MAX_INDIRECTIONS; ++i) {
+            node = replacements.get(node);
```
larsk21, line 229: `replacements` does always contain the node as key, right?

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
g4rgoil: Fair point

src/main/java/edu/kit/compiler/optimizations/common_subexpression/CommonSubexpressionElimination.java
g4rgoil: In theory, yes `node` _should_ never be `null` ...

larsk21: And if I loop for a couple of times, replacing `node` with `replacements.get(node)`, I will never find a `node != null` with `replacements.containsKey(node) == false`?

g4rgoil: Actually, this function is not correct as it is right now (see the newest commit). I'm really not sure what the check for `iro_Deleted` is supposed to do.

larsk21 (approved):
Ideally, we would fix the analysis first, but since this should give us an additional speedup, we should merge it now.

larsk21:
see #174

PR #109, opened 2022-01-05
Add Arithmetic Strength Reduction

g4rgoil:
Adds an optimization that implements strength reduction for a number of arithmetic operations. Also extends instruction selection to handle shifts and the logical and.

A list of all that is being replaced and what it is replaced with, can be found in the main source file. To summarize, multiplications by any power of two, as well as division and modulus operations with any constant divisor are removed.


N-Maas (approved):
Works well and actually provides a substantial performance improvement for some of the tests, probably thanks to the strength reduction for div.

I couldn't follow the details of the implementation, but I think this is due to the inherent complexity of the implemented algorithm. The structure of the code looks fine.

PR #110, opened 2022-01-05
Improved register allocation

N-Maas:
Implements a register allocation that generates actually performant code.

It contains the following improvements:
- Implementation of a linear scan register allocator that is used when optimizations are enabled. Specifically, the linear scan algorithm tries to find register assignments that avoid unnecessary spills or moves by calculating a register preference based on lifetime and usage
- Proper code generation for function calls by implementing a register permutation for register-passed arguments instead of using the stack
- Avoid unnecessary reloads of spilled registers within a single basic block by tracking the register states more precisely

Also includes a few minor fixes and cleanup.

Some further optimization possibilities that are not contained in this PR:
- reuse of stack slots for spilled registers
- allowing non-linear lifetimes (i.e. with holes) by creating cuts before and after calls, division/modulo and loops (this is non-trivial and would require some degree of control-flow awareness)


src/main/java/edu/kit/compiler/register_allocation/LinearScan.java
```diff
@@ -0,0 +1,427 @@
...
+            if (analysis.getLifetime(target).getBegin() == index) {
+                // allocate the new register
+                RegisterPreference preference = calculatePreference(target, state.getAssignment(), analysis);
+                state.allocateRegister(target, preference);
+            }
+        }
+    }
+
+    private static RegisterPreference calculatePreference(int vRegister, RegisterAssignment[] assignment,
+                                                          LifetimeAnalysis analysis) {
```
g4rgoil, line 117: More of a general point, there are a excessively long methods in this class, as well as in `ApplyAssignment`. Maybe some of these methods could be broken into multiple logic units, to improve readability and possibly find some opportunities for code-reuse.

g4rgoil (commented):
Generally looks good and passes every test case. Together with the improved instruction selection it seems to provide a significant performance uplift (runtimes of some of the larger test cases is more than halved).

src/main/java/edu/kit/compiler/register_allocation/LinearScan.java
N-Maas: I have looked into it and found a few minor things and one bigger block that could be refactored as logically separate units.

However, all methods are/were below or around 100 lines, so I wouldn't call them particularly long. I don't think introducing new methods only for the sake of reducing line counts is a good idea here.

g4rgoil: The changes you made look good. If there is no real opportunity for further code reuse, you're probably right and we can leave it as it is.

g4rgoil (approved):

PR #113, opened 2022-01-10
Improved handling of Div and Mod

pmehnert:
This PR seeks to improve our handling of division and modulus operations. The goal is to check for the edge case of `INT_MIN (/|%) -1` explicitly and to handle it separately. We will then be free to use 32-bit instead of 64-bit divisions.

- [x] Generate the required Firm graphs during transformation
- [x] Switch to double-word division during register allocation
- [x] Update strength reductions to work with the new graphs


g4rgoil:
There are currently two failing `ConstantOptimization` tests. It seems the constant propagation is not completed entirely in one pass. The tests are successful if the optimization is run twice. (@larsk21)

larsk21:
Yes, changing the Firm graph in the second step of constant propagation and the subsequent simplifications done by libfirm can lead to the constant propagation finding new constants to propagate.

N-Maas (approved):

PR #115, opened 2022-01-11
Remove unused imports and variables

larsk21:


src/test/java/edu/kit/compiler/transform/IRExpressionVisitorTest.java
```diff
@@ -46,7 +42,7 @@ private ExpressionNode getReturnExpressionFromProgram(String input) {
         for (ClassNode classNode: ast.getClasses()) {
             for (MethodNode.DynamicMethodNode m: classNode.getDynamicMethods()) {
                 var mapping = LocalVariableCounter.apply(m);
-                this.context = new TransformContext(
+                new TransformContext(
```
larsk21, line 45: Is this call necessary?

src/test/java/edu/kit/compiler/transform/TransformContextTest.java
```diff
@@ -50,7 +45,7 @@ private MethodNode initMethod(String input) {
         this.stringTable = lexer.getStringTable();
         this.typeMapper = new TypeMapper(namespaceMapper, lexer.getStringTable());
         for (ClassNode classNode: node.getClasses()) {
-            ClassEntry entry = typeMapper.getClassEntry(classNode);
+            typeMapper.getClassEntry(classNode);
```
larsk21, line 48: Is this call necessary?

src/test/java/edu/kit/compiler/transform/IRExpressionVisitorTest.java
larsk21: @0xNB

src/test/java/edu/kit/compiler/transform/TransformContextTest.java
larsk21: @N-Maas

N-Maas: No, I think this is a pure function anyway

src/test/java/edu/kit/compiler/transform/IRExpressionVisitorTest.java
N-Maas: I don't think so. Actually, the whole file seems quite unnecessary to me.

N-Maas (approved):

PR #116, opened 2022-01-11
Change detailed type analysis to use `ErrorHandler`

### LINKED ISSUE #60: Use `ErrorHandler` in detailed name/type analysis
It would improve the detailed name and type analysis, if we were able to report more than one error. The necessary changes include moving from exception throwing to collecting errors (easy), as well as taking into account that reporting an error would not end the method immediately, potentialy causing follow-up errors (more complicated).

larsk21:
closes #60 

fixed bugs:
- for a default binary expression with an incompatible type on the right side of the operator, the error message included the type of the expression on the left side
- if a method invocation had too many arguments, the nodes of the additional arguments were not visited

### DIFF
--- a/src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
+++ b/src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
@@ -24,34 +24,36 @@ import edu.kit.compiler.semantic.NamespaceMapper.ClassNamespace;
  * - builtin definitions (classes, fields, methods) are available
  * 
  * Postconditions:
- * - all used variables reference their declaration
+ * - all used variables reference their declaration *
  * - every field, method, statement and expression is correctly typed
  * - no variable is declared twice inside the same method
- * - every expression node contains a valid result type
+ * - every expression node contains a valid result type *
  * - static methods contain no reference to this
  * - void is not used as a field, parameter or local variable type, in a new
  * object or new array expression
- * - hasError is set in all nodes where an error occured
+ * - the error flag is set in all nodes where an error occured
  * - user defined data types of fields and methods (incl. parameters) are valid
  * - integer literal values are valid integer values (32-bit signed)
  * - calls to the standard library are identified
+ * 
+ * * unless the error flag is set for the node
  * 
  * Not checked:
  * - all code paths return a value
  * - left side of an assignment is an l-value
  */
-public class DetailedNameTypeAstVisitor implements AstVisitor<DataType> {
+public class DetailedNameTypeAstVisitor implements AstVisitor<Optional<DataType>> {
 
-    public static final DataType VoidType = new DataType(DataTypeClass.Void);
-
-    public DetailedNameTypeAstVisitor(NamespaceMapper namespaceMapper, StringTable stringTable) {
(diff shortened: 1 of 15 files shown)


larsk21:
Example with multiple errors:
```
class Test {
    public static void main(String[] args) {
        Test test = new Test();
        int x = test.f(5, true, test.g(17));
    }
    public boolean f(int x, int y) {
        return x == y;
    }
}
```

Output:
```
error: semantic check at line 4, column 22: wrong number of arguments
error: semantic check at line 4, column 22: argument 1 of type boolean does not match expected type int
error: semantic check at line 4, column 38: unknown method
error: semantic check at line 4, column 9: invalid assigment, variable type is int while expression type is boolean
```

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -125,210 +127,252 @@ public DataType visit(ClassNode classNode) {
...
-        return VoidType;
+        return Optional.empty();
     }
 
-    public DataType visit(BinaryExpressionNode binaryExpressionNode) {
-        DataType leftSideType = binaryExpressionNode.getLeftSide().accept(this);
-        DataType rightSideType = binaryExpressionNode.getRightSide().accept(this);
+    private Optional<DataType> applyResultType(ExpressionNode expressionNode, Optional<DataType> resultType_) {
+        resultType_.ifPresentOrElse(resultType -> {
+            expressionNode.setResultType(resultType);
```
g4rgoil, line 285: Some of your lambdas could be simplified using method references (in this case `expressionNode::setResultType`).

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
```diff
@@ -125,210 +127,252 @@ public DataType visit(ClassNode classNode) {
...
 
-                resultType = leftSideType;
+                resultType_ = leftSideType_;
                 break;
             case Equal:
             case NotEqual:
-                if (!leftSideType.isCompatibleTo(rightSideType)) {
-                    semanticError(binaryExpressionNode, "the two arguments of %s must have compatible types", operator);
-                }
+                leftSideType_.ifPresent(leftSideType -> rightSideType_.ifPresent(rightSideType -> {
```
g4rgoil, line 319: You use this double `ifPresent` construct quite a few times, you could consider a helper method for this to make it more readable (something like `ifBoth(Optional<S>, Optional<T>, BiFunction<S, T, R>)`).

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
larsk21: I'm aware and I, personally, like them better, I just made the experience that some developers find them hard to read.

src/main/java/edu/kit/compiler/semantic/DetailedNameTypeAstVisitor.java
larsk21: I'm unsure if it would be more readable, but I can change it if it helps.

g4rgoil (approved):
Looks good

PR #121, opened 2022-01-13
Handle Unknown nodes during instruction selection

g4rgoil:
The instruction selection must be able to deal with `Unknown` nodes. `Unknown` nodes may for example result from using the value of a local variable prior to assignment. As the definite assignment rules of Java don't apply in MiniJava, our compiler must be able to generate code for such programs.

### DIFF
--- a/src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
+++ b/src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
@@ -1,0 +1,60 @@
+package edu.kit.compiler.codegen.pattern;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import edu.kit.compiler.codegen.MatcherState;
+import edu.kit.compiler.codegen.Operand;
+import edu.kit.compiler.codegen.Util;
+import edu.kit.compiler.intermediate_lang.Instruction;
+import firm.bindings.binding_irnode.ir_opcode;
+import firm.nodes.Node;
+import lombok.RequiredArgsConstructor;
+
+public final class UnknownPattern implements Pattern<InstructionMatch> {
+
+    @Override
+    public InstructionMatch match(Node node, MatcherState matcher) {
+        if (node.getOpCode() == ir_opcode.iro_Unknown) {
+            var size = Util.getSize(node.getMode());
+            var targetRegister = matcher.getNewRegister(size);
+            return new UnknownMatch(node, targetRegister);
+        } else {
+            return InstructionMatch.none();
+        }
+    }
+
+    @RequiredArgsConstructor
+    private static final class UnknownMatch extends InstructionMatch.Basic {
+
+        private final Node node;
+        private final int targetRegister;
+
+        @Override
+        public Node getNode() {
+            return node;
+        }
+
+        @Override
+        public Optional<Integer> getTargetRegister() {
+            return Optional.of(targetRegister);
+        }
+
+        @Override
+        public Stream<Operand> getOperands() {
+            return Stream.empty();
+        }
+
+        @Override
+        public Stream<Node> getPredecessors() {
+            return Stream.empty();
+        }
+
+        @Override
+        public List<Instruction> getInstructions() {
+            return List.of(Instruction.newOp("nop", List.of(),
+                    Optional.empty(), targetRegister));
+        }
+    }
+}
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -67,6 +67,11 @@ jobs:
           pwd
           ./run_mjtests.sh compile-firm
 
+      - name: Run Compile only Tests
+        run: |
+          pwd
+          ./run_mjtests.sh compile-only
+
       - name: Run Compile Tests
         run: |
           pwd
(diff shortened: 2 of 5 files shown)


src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        @Override
+        public Stream<Node> getPredecessors() {
+            return Stream.empty();
+        }
+
+        @Override
+        public List<Instruction> getInstructions() {
+            return List.of(Instruction.newOp("nop", List.of(),
```
larsk21, line 56: Why is it necessary to output a `nop` instruction, instead of returning an empty list?

g4rgoil: It's a bit of a hack. I need to allocate a register but don't have anything to initialize it with. I could instead write some value to the register (it's UB so it doesn't matter which value).

larsk21: Hm, I see. Could we remove the `nop` at a later point? After all, it shouldn't be necessary to include it in the output file, right?

g4rgoil: We don't need the nop during execution, so it could be removed once register allocation is complete.

larsk21: Where would be the right place to do so? Should we add peephole optimizations now or is it enough to ignore `nop`s during the assembly file writing?

src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
```diff
@@ -0,0 +1,60 @@
...
+import firm.bindings.binding_irnode.ir_opcode;
+import firm.nodes.Node;
+import lombok.RequiredArgsConstructor;
+
+public final class UnknownPattern implements Pattern<InstructionMatch> {
+
+    @Override
+    public InstructionMatch match(Node node, MatcherState matcher) {
+        if (node.getOpCode() == ir_opcode.iro_Unknown) {
+            var size = Util.getSize(node.getMode());
```
larsk21, line 20: With the changes in #123, constant optimization will replace arithmetic / boolean operations with at least one unknown operand with a new `Unknown` node. In the following example, this leads to an `Unknown` node of mode `B` as the predecessor of the `Cond` node.

The problem: `Util.getSize` will fail for mode `B`.

Example:
```
class Test {
    public static void main(String[] args) {
        Test x;
        if (x == null) {}
    }
}
```
(similar to `CompareRefWithNull.valid.mj`)

src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
g4rgoil: For the assembly writer to remove instructions might be a bit unexpected, a peephole optimization would probably be the cleaner solution.

larsk21: Of course, but implementing this in a clean way would require some framework for peephole optimizations. That seems like an overkill for only removing `nop`s. Maybe we can target this later.

g4rgoil: You're right, if we want to remove `nop`s we could do it in the assembly writer for now.

larsk21: Or just ignore them for now ;) After all, having some `nop`s doesn't hurt us.

src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
g4rgoil: Hm, that is a problem,  I can't (and it wouldn't make sense to) allocate a register for a Firm boolean. We could replace `Cond(Unknown)` with an unconditional jump to either of the control flow successors.

larsk21: That would be a valid solution since `Cond(Unknown)` produces undefined behavior. Where would we implement this though?

g4rgoil: It needs to run for `-O0` as well, so currently either during constant propagation or arithmetic identities.

larsk21: My first thought is that it's neither a constant propagation nor an arithmetic identity, it's just resolving undefined behavior. But if it needs to be in one of the two, I guess it would fit better in constant propagation.

g4rgoil: We could obviously introduce an entirely new optimization, but that's another graph pass each iteration. Also I don't think it's entire misplaced in constant propagation. You're already propagating `Unknown`s for other kinds of nodes, so why not do it for `Cond` as well.

larsk21: Alright, I will include it in #123.

larsk21 (approved):

PR #123, opened 2022-01-13
Fix handling of `Unknown` nodes in constant optimization

larsk21:


g4rgoil:
Will this optimize something like `Add(anyNode, Unknown)`?

larsk21:
> Will this optimize something like `Add(anyNode, Unknown)`?

Yes, this is demonstrated in the added uni test. `y + 5` with an unknown `y` is changed to an `Unknown` node.

src/main/java/edu/kit/compiler/optimizations/ConstantOptimization.java
```diff
@@ -126,28 +133,50 @@ private boolean transformMemoryProj(Proj node) {
...
+
+            if (predValue.isUnknown()) {
+                predValue = TargetValueLatticeElement.constant(undefinedCondStrategy.chooseCondValue((Cond)pred));
+                nodeValues.put(pred, predValue);
+            }
+
+            if (predValue.isConstant()) {
+                if (
+                    (node.getNum() == pn_Cond.pn_Cond_false.val && predValue.getValue().equals(TargetValue.getBFalse())) ||
+                    (node.getNum() == pn_Cond.pn_Cond_true.val && predValue.getValue().equals(TargetValue.getBTrue()))
```
g4rgoil, line 147: Is there a reason you're not using `Cond.pn{True,False}`?

g4rgoil (approved):
Seems to fully resolve the issue of `Unknown` nodes. Maybe we should note somewhere that our instruction selection is not guaranteed to work for graphs where this optimization has not been applied.

src/main/java/edu/kit/compiler/optimizations/ConstantOptimization.java
larsk21: No, I just copied it from where I used this before :D

larsk21:
> Seems to fully resolve the issue of `Unknown` nodes. Maybe we should note somewhere that our instruction selection is not guaranteed to work for graphs where this optimization has not been applied.

I added some documentation and a safety-check (which does not resolve the problem though). Maybe we could add a pattern for `Cond` nodes with an `Unknown` node as the selector, just to be safe.

g4rgoil:
> I added some documentation and a safety-check (which does not resolve the problem though). Maybe we could add a pattern for `Cond` nodes with an `Unknown` node as the selector, just to be safe.

How should a `Cond` with `Unknown` selector be handled during instruction selection in your opinion? Just generate an unconditional jump?

larsk21:
> > I added some documentation and a safety-check (which does not resolve the problem though). Maybe we could add a pattern for `Cond` nodes with an `Unknown` node as the selector, just to be safe.
> 
> How should a `Cond` with `Unknown` selector be handled during instruction selection in your opinion? Just generate an unconditional jump?

For example, anything that generates valid code would be acceptable. This way the instruction selection would be independent of the optimizations.

g4rgoil:
> > > I added some documentation and a safety-check (which does not resolve the problem though). Maybe we could add a pattern for `Cond` nodes with an `Unknown` node as the selector, just to be safe.
> > 
> > 
> > How should a `Cond` with `Unknown` selector be handled during instruction selection in your opinion? Just generate an unconditional jump?
> 
> For example, anything that generates valid code would be acceptable. This way the instruction selection would be independent of the optimizations.

Ok, I can implement a pattern for that.

PR #124, opened 2022-01-13
Improve usage of addressing modes

g4rgoil:
This PR implements two small transformations to the Firm graph. When combined these lead to better usage of addressing modes in certain situations. For example the following function will translate to something like `movl 16(%rsi,%rdx,4), %eax`.

```java
public int foo(int[] arr, int i) {
    return arr[i + 4];
}
```


N-Maas (approved):
Looks good to me :+1:

PR #125, opened 2022-01-13
Assume that array indizes are always positive

N-Maas:
Negative indizes are undefined behavior in MiniJava. We can use this to avoid emitting an unnecessary `movslq`.

Note: This does not work if the conversion is propagated to operands, because operands on their own might still be negative.

### DIFF
--- a/src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
+++ b/src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
@@ -251,19 +251,24 @@ public final class ArithmeticIdentitiesOptimization implements Optimization {
 
         @Override
         public void visit(Conv node) {
+            // this works in combination with distributive transformation of
+            // multiplications to better use x86 addressing modes
             if (node.getMode().equals(Mode.getLs())
                     && node.getOp().getOpCode() == ir_opcode.iro_Add
                     && isConst(node.getOp().getPred(1))) {
                 // (x + c):Ls --> x:Ls + c:Ls
-                // this works in combination with distributive transformation of
-                // multiplications to better use x86 addressing modes
-                var block = node.getBlock();
-                var opNode = graph.newConv(block, node.getOp().getPred(0), Mode.getLs());
-                var offset = ((Const) node.getOp().getPred(1)).getTarval();
-                var offsetNode = graph.newConst(offset.convertTo(Mode.getLs()));
-                var addNode = graph.newAdd(block, opNode, offsetNode);
-                exchange(node, addNode);
+                exchangeConv(node, node.getOp().getPred(0), node.getOp().getPred(1));
+            } else if (node.getMode().equals(Mode.getLs())
+                    && node.getOp().getOpCode() == ir_opcode.iro_Conv) {
+                Conv luNode = (Conv) node.getOp();
+                if (luNode.getOp().getOpCode() == ir_opcode.iro_Add
+                        && isConst(luNode.getOp().getPred(1))) {
+                    // (x + c):Lu:Ls --> x:Ls + c:Ls
+                    exchangeConv(node, luNode.getOp().getPred(0), luNode.getOp().getPred(1));
+                }
             }
+            // Note: We can not apply the same optimization for Lu as for Ls,
+            // because this could change the program semantics
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -251,19 +251,24 @@ public void visit(Mod node) {
...
-                var addNode = graph.newAdd(block, opNode, offsetNode);
-                exchange(node, addNode);
+                exchangeConv(node, node.getOp().getPred(0), node.getOp().getPred(1));
+            } else if (node.getMode().equals(Mode.getLs())
+                    && node.getOp().getOpCode() == ir_opcode.iro_Conv) {
+                Conv luNode = (Conv) node.getOp();
+                if (luNode.getOp().getOpCode() == ir_opcode.iro_Add
+                        && isConst(luNode.getOp().getPred(1))) {
+                    // (x + c):Lu:Ls --> x:Ls + c:Ls
+                    exchangeConv(node, luNode.getOp().getPred(0), luNode.getOp().getPred(1));
```
g4rgoil, line 267: Do I understand this right that this transformation is only correct, because we assume that the result of the addition is always non-negative? And if we used Lu for anything other than array indexing this would change the semantics of the program.

N-Maas: No, this transformation should always be correct. It exchanges a possibly lossy conversion (? -> Lu -> Ls) with a lossless conversion (? -> Ls).

Assuming the values fit in `Ls`, I can't see how this could change semantics.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
```diff
@@ -338,6 +343,18 @@ private void foldMultiplication(Mul node) {
             }
         }
 
+        /**
+         * Exchanges a conv node which operand is an add with a constant as rhs.
+         */
+        private void exchangeConv(Node node, Node opNode, Node offsetNode) {
+            var block = node.getBlock();
+            var op = graph.newConv(block, opNode, Mode.getLs());
```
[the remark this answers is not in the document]
N-Maas: I have added an assertion, but I don't think there is an assumption other than that the Mode of `node` is `Ls`.

src/main/java/edu/kit/compiler/optimizations/ArithmeticIdentitiesOptimization.java
N-Maas: Okay, except you want to use an `Lu` conversion to somehow reinterpret a negative number as a positive number? But this sounds like a really bad idea to me and I would argue that this should be Undefined Behavior.

g4rgoil: Yes, that is what I was thinking about, casting a negative `Is` to `Lu`.

So, I've had a look in the Firm documentation. It seems what you're doing here is not entirely legal (it seems I've made the same mistake as well), see [tarval_convert_to()](https://pp.ipd.kit.edu/firm/api_1.22/a00151.html#gad9849e2e3a11797c9f0b4f08341b940a) and [smaller_mode()](https://pp.ipd.kit.edu/firm/api_1.22/a00141.html#ga5a593076885591ab7a7cf26f8b127e65).
I think we need to use `BitCast` here (i.e. `BitCast(Is -> Iu) -> Ls`).

N-Maas: I wanted to use a `BitCast` at first, but the verifier doesn't allow it. Here is the comment from the firm source code:

```
/* Note: This constraint is currently strict as you can use Conv
* for the other cases and we want to avoid having 2 nodes representing the
* same operation. We might loosen this constraint in the future. */
```

Bitcast is only allowed if the two modes use different arithmetic (and arithmetic means something like 2s complement, 1s complement or, respective, int/float here).

Anyway, there is no reason why our semantics shouldn't be stricter than that of firm. They already are at several points in the backend.

g4rgoil: Ok, it seems I got fooled by by the documentation of `convertTo`. In that case I think we can merge this as is.

g4rgoil (approved):

PR #126, opened 2022-01-14
Handle `Unknown` control flow in instruction selection

g4rgoil:
Note: Also includes some general cleanup in `InstructionSelection`.


src/main/java/edu/kit/compiler/codegen/pattern/UnknownPattern.java
```diff
@@ -22,10 +21,15 @@
```
larsk21, line 21: ``except with Mode `b` ``

This should probably be removed, since `Unknown` nodes with Mode `b` are matched (although the call only returns a placeholder).

larsk21 (approved):

PR #127, opened 2022-01-15
Add call graph and global optimizations

g4rgoil:
This PR introduces an interface for global optimizations (i.e. optimizations that work on the entire program instead of a single graph). A call graph is also implemented to query calling-realtionships between functions and to detect recursions.

### DIFF
--- a/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
+++ b/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
@@ -37,7 +37,7 @@ import edu.kit.compiler.logger.Logger.Verbosity;
 import edu.kit.compiler.optimizations.ArithmeticIdentitiesOptimization;
 import edu.kit.compiler.optimizations.ArithmeticReplacementOptimization;
 import edu.kit.compiler.optimizations.ConstantOptimization;
-import edu.kit.compiler.optimizations.Optimization;
+import edu.kit.compiler.optimizations.Optimizer;
 import edu.kit.compiler.parser.Parser;
 import edu.kit.compiler.parser.PrettyPrintAstVisitor;
 import edu.kit.compiler.semantic.DetailedNameTypeAstVisitor;
@@ -181,10 +181,10 @@ public class JavaEasyCompiler {
      * @param logger the logger
      * @return Ok or an according error
      */
-    private static Result compileFirm(String filePath, Logger logger, Iterable<Optimization> optimizations,
-                                        DebugFlags debugFlags) {
+    private static Result compileFirm(String filePath, Logger logger, Optimizer optimizer,
+                                      DebugFlags debugFlags) {
         try {
-            createOptimizedIR(filePath, logger, optimizations, debugFlags);
+            createOptimizedIR(filePath, logger, optimizer, debugFlags);
 
             var sourceFile = new File(filePath).getName();
             var assemblyFile = sourceFile + ".s";
@@ -198,13 +198,13 @@ public class JavaEasyCompiler {
             }
 
             logger.info("compiling program: 'gcc \"%s\" \"%s\"'", assemblyFile, stdLibrary);
-            var process = Runtime.getRuntime().exec(new String[]{ "gcc", assemblyFile, stdLibrary });
+            var process = Runtime.getRuntime().exec(new String[] { "gcc", assemblyFile, stdLibrary });
 
             if (process.waitFor() != 0) {
                 logger.error("gcc failed with exit code %s", process.exitValue());
                 return Result.GccError;
             }
-            
+
             return Result.Ok;
         } catch (CompilerException e) {
             logger.withName(e.getCompilerStage().orElse(null)).exception(e);
@@ -228,10 +228,10 @@ public class JavaEasyCompiler {
      * @param logger the logger
      * @return Ok or an according error
      */
-    private static Result compile(String filePath, Logger logger, Iterable<Optimization> optimizations,
+    private static Result compile(String filePath, Logger logger, Optimizer optimizer,
                                   RegisterAllocator allocator, DebugFlags debugFlags) {
         try {
-            createOptimizedIR(filePath, logger, optimizations, debugFlags);
+            createOptimizedIR(filePath, logger, optimizer, debugFlags);
 
             PatternCollection coll = new PatternCollection();
             List<FunctionInstructions> functions = new ArrayList<>();
@@ -265,7 +265,7 @@ public class JavaEasyCompiler {
             }
 
             logger.info("compiling program: 'gcc \"%s\" \"%s\"'", assemblyFile, stdLibrary);
-            var process = Runtime.getRuntime().exec(new String[]{ "gcc", assemblyFile, stdLibrary });
+            var process = Runtime.getRuntime().exec(new String[] { "gcc", assemblyFile, stdLibrary });
 
             if (process.waitFor() != 0) {
                 logger.error("gcc failed with exit code %s", process.exitValue());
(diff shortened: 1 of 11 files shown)


src/main/java/edu/kit/compiler/optimizations/Optimizer.java
```diff
@@ -0,0 +1,106 @@
...
+     * Returns a set of all graphs in the program.
+     */
+    private HashSet<Graph> getAllGraphs() {
+        return StreamSupport
+                .stream(Program.getGraphs().spliterator(), false)
+                .collect(Collectors.toCollection(HashSet::new));
+    }
+
+    private void dumpGraphs(String prefix) {
+        if (dumpGraphs) {
```
N-Maas, line 100: A bit of an unfortunate name collision. Also, the method name sounds as if the graphs are dumped unconditionally.

g4rgoil: Good point, I'll change that to something more sensible.

N-Maas (commented):
The code generally looks good.

However, I have some high-level remarks:
- Do you have some specific examples of global optimizations in mind? At the moment, I can only think of dead code elimination and for this, we probably don't need a general framework. Inlining is not a really good canditate for a global optimization, it would converge much quicker if applied as a local optimization. (Of course, the functions would need to be optimized in bottom-up order of the callgraph.) So, because the runtime of the compiler is not really important, we could probably do it as a global optimization, but it is not my preferred approach.
- At least for some analysis with regards to inlining, it would be helpful if the call graph also represents the number of occurences of a specific callee within a caller. You could probably use edge weights for this.

g4rgoil:
> The code generally looks good.
> 
> However, I have some high-level remarks:
> 
>     * Do you have some specific examples of global optimizations in mind? At the moment, I can only think of dead code elimination and for this, we probably don't need a general framework. Inlining is not a really good canditate for a global optimization, it would converge much quicker if applied as a local optimization. (Of course, the functions would need to be optimized in bottom-up order of the callgraph.) So, because the runtime of the compiler is not really important, we could probably do it as a global optimization, but it is not my preferred approach.
> 
>     * At least for some analysis with regards to inlining, it would be helpful if the call graph also represents the number of occurences of a specific callee within a caller. You could probably use edge weights for this.

Regarding your first point, I would use global optimizations in cases where we need non-local analysis of the program. For example in the case of pure function optimization, you need to analyze the entire program and then do something with that information for each graph. Our interface for local optimizations doesn't give you any way to do that efficiently.

On the subject of how to store the number of distinct call-sites, the call graph was originally a multigraph (or pseudograph to be precise) where each edge represents exactly one invocation. I could return to that (though it makes the graph less efficient) or I could use weights if you preferred (note that the weights would have to be `double` values).

N-Maas:
> Regarding your first point, I would use global optimizations in cases where we need non-local analysis of the program. For example in the case of pure function optimization, you need to analyze the entire program and then do something with that information for each graph. Our interface for local optimizations doesn't give you any way to do that efficiently.

Okay, for pure function optimization this is fine. But I still want to use the call graph to apply the local optimizations in bottom-up order.

> On the subject of how to store the number of distinct call-sites, the call graph was originally a multigraph (or pseudograph to be precise) where each edge represents exactly one invocation. I could return to that (though it makes the graph less efficient) or I could use weights if you preferred (note that the weights would have to be `double` values).

I can work with `double` without problems :+1:

g4rgoil:
> Okay, for pure function optimization this is fine. But I still want to use the call graph to apply the local optimizations in bottom-up order.

Just to clarify, you would implement the inliner as local optimization? So, the interface would need a parameter for the call graph and it would need to be updated during `Optimizer#optimizeLocal()`.

N-Maas:
> > Okay, for pure function optimization this is fine. But I still want to use the call graph to apply the local optimizations in bottom-up order.
> 
> Just to clarify, you would implement the inliner as local optimization? So, the interface would need a parameter for the call graph and it would need to be updated during `Optimizer#optimizeLocal()`.
 
Yes. I think the best approach is to add some kind of `OptimizationContext` to the interface, although this doesn't need to happen in this PR. Then, we probably want to update the context (including the call graph) every time local optimizations for a function are done and after global optimization passes.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -331,9 +330,8 @@ private static ProgramNode createAttributedAst(
      * @param logger the logger
      * @return Ok or an according error
      */
-    private static void createOptimizedIR(String filePath, Logger logger,
-                                            Iterable<Optimization> optimizations,
-                                            DebugFlags debugFlags) throws IOException {
+    private static void createOptimizedIR(String filePath, Logger logger, Optimizer optimizer,
+                                          DebugFlags debugFlags) throws IOException {
```
larsk21, line 334: `debugFlags` is no longer needed here, right?

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -385,23 +363,22 @@ public static void main(String[] args) {
...
                 allocator = new DumbAllocator();
                 break;
             case Level1:
-                optimizations = Arrays.asList(
+                optimizer = new Optimizer(List.of(), List.of(
                     new ConstantOptimization(),
                     new ArithmeticIdentitiesOptimization(),
                     new ArithmeticReplacementOptimization()
-                );
+                ), debugFlags.isDumpGraphs());
```
larsk21, line 381: I would prefer passing the entire `debugFlags` to the optimizer. That would make it easier to extend it, in case we want to use additional debug flags in the future. The usage in `Optimizer` would change only slightly to `debugFlags.isDumpGraphs`.

src/main/java/edu/kit/compiler/optimizations/Optimizer.java
```diff
@@ -0,0 +1,106 @@
...
+            var callGraph = CallGraph.create();
+
+            // ? maybe only run global opts once per iteration
+            Set<Graph> newChanges;
+            do {
+                newChanges = optimizeGlobal(callGraph);
+                changeSet.addAll(newChanges);
+            } while (!newChanges.isEmpty());
+
+            hasChanged = optimizeLocal(changeSet);
```
larsk21, line 46: Is there a reason why we run the global optimizations first? I imagine local optimizations are more efficient, given that they have less information to evaluate (only the local graph). So we could first try to optimize as much as possible locally and then apply global optimizations afterwards (in rounds, of course).

src/main/java/edu/kit/compiler/optimizations/CallGraph.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /**
+     * Returns true if there may exist a recursion if caller calls callee,
+     * i.e. there exists some path through the call graph such that caller
+     * (in-)directly calls itself.
+     */
+    public boolean existsRecursion(Entity caller, Entity callee) {
+        // ? what should this fn do if caller does not call callee
+        // ? currently returns false, maybe throw exception instead
```
larsk21, line 82: I think returning `false` is alright. If the user of this function is interested in knowing whether a call exists at all, they should call `existsCall` in addition.

src/main/java/edu/kit/compiler/optimizations/CallGraph.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    @RequiredArgsConstructor
+    @AllArgsConstructor
+    private static final class Visitor extends NodeVisitor.Default {
+
+        private Entity caller;
+        private final Graph<Entity, DefaultEdge> graph;
+
+        public static CallGraph create() {
```
larsk21, line 195: Are two `create` functions necessary here? You could simply call `Visitor.create(Program.getGraphs())` in `CallGraph::create` and have a cleaner interface for `Visitor`.

src/main/java/edu/kit/compiler/optimizations/Optimizer.java
g4rgoil: I've swapped the order of local and global optimizations, there was no deeper reasoning behind the old order. You're probably right that optimizing locally as far as possible first, may be more efficient.
Regarding your point about local optimizations having less information, that may be subject to change (see discussion thread).

N-Maas (approved):
Looks good to me

PR #130, opened 2022-01-20
Add `lea` pattern to instruction selection

g4rgoil:
This PR adds functionality to allow for the use of load effective address during instruction selection. This just reuses the memory patterns used to match `Load` nodes. Most additions of registers, additions with constants, as well as a limited number of multiplications (those with 2, 4, or 8) can now be expressed using `lea`. 

Note: There is still room for improvement here. For example multiplications by 3, 5, and 9 could also be expressed using `lea`.

### DIFF
--- a/mjtest-files/compile/pattern/LoadEffective.java
+++ b/mjtest-files/compile/pattern/LoadEffective.java
@@ -1,0 +1,79 @@
+class Main {
+    public static void main(String[] args) {
+        Main m = new Main();
+
+        System.out.println(m.baseOffset(0));
+        System.out.println(m.baseOffset(128));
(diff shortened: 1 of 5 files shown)


larsk21 (approved):
Looks good to me and generates some nice assembler code!

The only thing I wondered: Is `lea (,_,8), _` really faster than `shl _,$3` (plus maybe the moves from and to `%eax`)?

g4rgoil:
> Looks good to me and generates some nice assembler code!
> 
> The only thing I wondered: Is `lea (,_,8), _` really faster than `shl _,$3` (plus maybe the moves from and to `%eax`)?

I would assume that it is at least as fast. I generally use what gcc and clang do as guidance.

PR #131, opened 2022-01-20
Add linear blocks optimization

larsk21:
See `LinearBlockVisitor::visit` for an explanation when blocks can be joined.

### DIFF
--- a/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
+++ b/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
@@ -37,6 +37,7 @@ import edu.kit.compiler.logger.Logger.Verbosity;
 import edu.kit.compiler.optimizations.ArithmeticIdentitiesOptimization;
 import edu.kit.compiler.optimizations.ArithmeticReplacementOptimization;
 import edu.kit.compiler.optimizations.ConstantOptimization;
+import edu.kit.compiler.optimizations.LinearBlocksOptimization;
 import edu.kit.compiler.optimizations.Optimizer;
 import edu.kit.compiler.parser.Parser;
 import edu.kit.compiler.parser.PrettyPrintAstVisitor;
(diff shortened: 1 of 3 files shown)


larsk21:
Most of the mjtests run fine, but there is one with a weird bug. Compiling and running `compile/CellularAutomaton.mj` will result in an inverted output grid (where the program should output the character `'` a `#` is output and vice versa), otherwise it runs fine.

Even more weird:
- when compiled with the firm backend, the bug is not present
- when compiled with the advanced register allocator (even without other optimizations), the bug is present
- when compiled with the dumb allocator (without other optimizations), a segmentation fault occurs, originating in an `movq (_, _, 8), _` instruction in the method `CellularAutomaton::run`

Looking at the graphs, the optimization doesn't change much, only a few empty blocks are skipped. This seems to be valid, since it compiles with the firm backend, but I have no idea where the error might come from.

N-Maas:
Interesting, I will try to look into it soon.

g4rgoil:
The optimized graph for `run` contains a block with two control flow predecessors pointing to nodes in the same block (just different results of a condition). I don't have any evidence to suggest it is related to this bug in any way, but it could cause problems (for example when trying to insert blocks for critical edge resolution).

image

larsk21:
Very interesting! But having a block with multiple equal predecessors is still valid in a Firm graph, isn't it?

g4rgoil:
> Very interesting! But having a block with multiple equal predecessors is still valid in a Firm graph, isn't it?

Yes, as far as I know that is absolutely valid in Firm. In some cases you can optimize it to an unconditional jump, but in general (for example in this case) you can't because of `Phi`s.

larsk21:
> Yes, as far as I know that is absolutely valid in Firm. In some cases you can optimize it to an unconditional jump, but in general (for example in this case) you can't because of `Phi`s.

Yes, that's also done in this optimization.

So we will fix this in the PhiResolver? Do you know who implemented it?

g4rgoil:
> So we will fix this in the PhiResolver? Do you know who implemented it?

Yeah, this needs to be fixed in `PhiResolver`, it may also require changes to `PermutationSolver`. I believe it was implemented by @N-Maas originally.

N-Maas (approved):
Looks good to me and works well.

PR #134, opened 2022-01-21
Handle identical predecessor block in phi resolver

N-Maas:


larsk21 (approved):
Looks good to me and resolves the problem mentioned in #131.

PR #139, opened 2022-01-24
Fix handling of new `Jmp` in linear blocks optimization

larsk21:


N-Maas (commented):
Looks good, but could you maybe also add the failing mj-test to our tests in order to avoid regressions?

larsk21:
> Looks good, but could you maybe also add the failing mj-test to our tests in order to avoid regressions?

In which folder do I put it to just compile it? If I put it in `compile`, it seems to execute it afterwards. (`Command 'java BooleanArrayInfiniteLoop' timed out`)

N-Maas (approved):

PR #140, opened 2022-01-24
Fix lifetime analysis regarding intermangled loops

N-Maas:
Previously, the lifetime analysis assumed that for nested loops, the blocks are ordered in a way such that the blocks of the outer loop enclose the blocks of the inner loop, thus allowing the lifetime analysis to find loops based on the back jump at the end of the loop.

However, this ordering is not always possible anymore with the linear blocks optimization, because there might be e.g. a direct jump from the header of the inner loop to the header of the outer loop.

Therefore, I have changed the lifetime analysis to use loop depths instead, which are set by the analysis done for the block ordering (`ReversePostfixOrder`). Additionally, `ReversePostfixOrder` must now ensure that the loop depths are "flat", i.e. there can be no hole with lower depth in a loop, which could be created e.g. by a return within the loop.

Note that the ordering of the blocks is unchanged by this PR, the only visible change from outside is that the lifetime analysis can now determine the correct lifetimes in more (hopefully, all) cases.

### DIFF
--- a/mjtest-files/compile/LifetimeError.mj
+++ b/mjtest-files/compile/LifetimeError.mj
@@ -1,0 +1,9 @@
+class A {
+    public boolean x;
+
+    public static void main(String[] args) {
+        A a = new A();
+        a.x = false;
+        while(a.x) while(!a.x) while(a.x) System.out.println(0);
+    }
+}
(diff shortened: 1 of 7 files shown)


g4rgoil (approved):
It certainly does seem to fix the issue, even though I'm not entirely sure I understand the specifics of your fix.

N-Maas:
> It certainly does seem to fix the issue, even though I'm not entirely sure I understand the specifics of your fix.

Yeah, the block layout algorithm is actually a bit complicated :smile:

PR #141, opened 2022-01-24
fix: Correctly replace const calls

g4rgoil:
Previously, the tuples used to remove calls to constant functions would directly point to the call. This is wrong, as it omits the required projection to get the result tuple of the call.


larsk21 (approved):
Seems to fix the problem

PR #143, opened 2022-01-26
Add assembly optimizations

g4rgoil:
This PR introduces a basic framework for assembly optimizations and uses it to implement the following two optimizations.

1. Jump Inversion (I don't know if that is the correct name) uses fall-through to optimize conditional jumps followed by unconditional ones where possible. For example, the following list of instruction
    ```asm
    jnz .L1
    jmp .L2
    .L1:
    ```
    is simplified to
    ```asm
    jz .L2
    .L1:
    ```

2. Remove all `nop`s, which may have been introduced to deal with `Unknown` nodes during instruction selection.

### DIFF
--- a/src/main/java/edu/kit/compiler/assembly/JumpInversion.java
+++ b/src/main/java/edu/kit/compiler/assembly/JumpInversion.java
@@ -1,0 +1,160 @@
+package edu.kit.compiler.assembly;
+
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import edu.kit.compiler.io.BufferedLookaheadIterator;
+import edu.kit.compiler.io.LookaheadIterator;
+import lombok.Getter;
+
+public final class JumpInversion implements Iterator<String> {
+
+    private final LookaheadIterator<String> input;
+
+    public JumpInversion(FunctionInstructions function) {
+        this.input = new BufferedLookaheadIterator<>(
+                function.getInstructions().iterator());
+    }
+
+    @Override
+    public boolean hasNext() {
+        return input.has();
+    }
+
+    @Override
+    public String next() {
+        if (input.has(2)) {
+            return parseConditional();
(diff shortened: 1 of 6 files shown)


src/main/java/edu/kit/compiler/assembly/JumpInversion.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+
+        private String parseLabel(Jump trueJump, String trueLabel, String falseLabel) {
+            assert input.has(2);
+
+            var instruction = input.get(2);
+            if (instruction.startsWith(trueLabel) && instruction.endsWith(":")) {
+                input.next(2);
+                return trueJump.getInverse().getInstruction() + " " + falseLabel;
```
N-Maas, line 83: This merges the instruction and the label into one line, right? This is probably not wrong, considering the way we handle the assembly, but it is rather ugly...

src/main/java/edu/kit/compiler/io/LookaheadIterator.java
```diff
@@ -37,4 +37,12 @@
      */
     void next(int steps);
 
+    /**
+     * Return the next element and advance the iterator.
+     */
+    default T getNext() {
```
N-Maas, line 43: This reads as if the method does not mutate the iterator.

N-Maas (commented):
We probably won't add many assembly optimizations, so this is not overly important, but the iterator-based interface doesn't seem to make it easy to implement such optimizations. Specifically, the control flow in the `JumpInversion` optimization is not that easy to follow and the optimization somewhat needs to merge two lines into one, because it can't easily return them as separate lines.

I wonder whether a better interface would be to give the optimization a fixed number of lines (probably 3, or configurable) and let it either return nothing if the lines should not be changed or a replacement otherwise. I think this would simplify optimizations that consider more than one line. (And we probably only want peephole optimizations here anyway.)

N-Maas (approved):
Your changes look good to me :+1:

PR #144, opened 2022-01-28
Minor register allocation improvements

N-Maas:
Adds two minor improvements to the register allocator:
- The allocator now only aligns the stack pointer to 16 byte if the call is a standard library call
- When spilling a register, the allocator is now able to reuse stack slots and also able to use the stack slot of a stack-passed argument. Further, the PR adds a small refinement to the heuristic for selecting the spilled register.

(Note: The second improvement applies quite seldomly, as we are rarely spilling registers anyway.)


g4rgoil (approved):
The implemented improvements make sense and seem to work just fine. Comparing `LdNames` is not ideal, but should work for our purposes.

PR #145, opened 2022-01-29
Add loop invariant optimization

larsk21:
Adds an optimization that moves loop-invariant nodes in front of the loop (to the deepest common dominator of the loop entry point predecessors). In nested loops, nodes are moved out as far as possible. Currently, only data nodes are moved (except Phis), but const/pure (? @g4rgoil) functions could be added, if applicable.

### DIFF
--- a/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
+++ b/src/main/java/edu/kit/compiler/JavaEasyCompiler.java
@@ -35,6 +35,7 @@ import edu.kit.compiler.optimizations.ArithmeticIdentitiesOptimization;
 import edu.kit.compiler.optimizations.ArithmeticReplacementOptimization;
 import edu.kit.compiler.optimizations.ConstantOptimization;
 import edu.kit.compiler.optimizations.LinearBlocksOptimization;
+import edu.kit.compiler.optimizations.LoopInvariantOptimization;
 import edu.kit.compiler.optimizations.Optimizer;
 import edu.kit.compiler.optimizations.PureFunctionOptimization;
 import edu.kit.compiler.parser.Parser;
@@ -378,7 +379,8 @@ public class JavaEasyCompiler {
                     new ArithmeticReplacementOptimization(),
                     new LinearBlocksOptimization(),
                     new InliningOptimization(),
-                    new PureFunctionOptimization()
+                    new PureFunctionOptimization(),
+                    new LoopInvariantOptimization()
                 ), debugFlags);
                 allocator = new LinearScan();
                 break;
--- a/src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
+++ b/src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
@@ -17,7 +17,8 @@ import firm.nodes.Node;
  * loop entry point. Nodes invariant to inner blocks are moved as far out as
  * possible.
  * 
- * Control flow, memory and Phi nodes are not affected.
+ * This analysis respects the pin state of nodes. In addition, control flow and
+ * Phi nodes are never moved.
  */
 public class LoopInvariantOptimization implements Optimization.Local {
 
--- a/src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
+++ b/src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
@@ -1,11 +1,13 @@
 package edu.kit.compiler.optimizations;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import edu.kit.compiler.io.CommonUtil;
 import edu.kit.compiler.optimizations.analysis.LoopInvariantAnalysis;
+import edu.kit.compiler.optimizations.loop_invariant.MoveInvariantStrategies;
 
 import firm.Graph;
 import firm.nodes.Block;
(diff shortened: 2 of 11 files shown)


g4rgoil:
> Currently, only data nodes are moved (except Phis), but const/pure (? @g4rgoil) functions could be added, if applicable.

You could use the analysis provided by the `OptimizationState`, however for your usecase it may be easier to just check whether a `Call` is pinned or not. _Pure_ functions are unpinned and can be moved freely, obviously keeping in mind the memory dependency. Beware that _Call_s may use  the `NoMem` node.

larsk21:
Okay, so `const` calls can be moved freely, because they don't have a memory dependency.

`pure` calls are more complicated though. Even when unpinned by the pure function optimization, they still have a memory dependency on the memory phi of the loop. That means I can't possibly move them in front of the loop, which is by design, given there could be other calls in the loop modifying the memory, on which the pure function still depends. Could the pure function optimization solve this by parallelizing memory dependencies of consecutive calls to pure functions? I.e. if I have a non-pure function call followed by two pure function calls, could the memory dependency of both pure function calls point to the memory result of the non-pure function? If this concept could be extended to loops, the loop phi would become unnecessary if all function calls in the loop are pure. Only then this optimization could start moving pure function calls, if I understand this right.

g4rgoil:
> Okay, so `const` calls can be moved freely, because they don't have a memory dependency.
> 
> `pure` calls are more complicated though. Even when unpinned by the pure function optimization, they still have a memory dependency on the memory phi of the loop. That means I can't possibly move them in front of the loop, which is by design, given there could be other calls in the loop modifying the memory, on which the pure function still depends. Could the pure function optimization solve this by parallelizing memory dependencies of consecutive calls to pure functions? I.e. if I have a non-pure function call followed by two pure function calls, could the memory dependency of both pure function calls point to the memory result of the non-pure function? If this concept could be extended to loops, the loop phi would become unnecessary if all function calls in the loop are pure. Only then this optimization could start moving pure function calls, if I understand this right.

What you're describing is definitely possible, it should however be done as part of a Load/Store optimization, as calls to _pure_ functions can essentially be seen as a load. As far as I know, @0xNB is developing such an optimization.

larsk21:
Okay, I will change the optimization to move calls as far as possible given their memory dependency then. If we add additional optimizations for that, the optimization should be able to pick it up.

larsk21:
But as it is now, the pure function optimization leaves memory dependencies of subsequent function calls to the memory result of a pure function unchanged, even though the pure function does not change the memory, right?

g4rgoil:
> But as it is now, the pure function optimization leaves memory dependencies of subsequent function calls to the memory result of a pure function unchanged, even though the pure function does not change the memory, right?

Yes, that is correct. The pure function optimization does not change the memory chain, with the exception of _const_ calls. Changing memory dependencies really only makes sense, if you do a full analysis. Handling a few special cases concerning pure functions is not really worth in my opinion, if the Load/Store optimization needs to consider function calls anyway.

N-Maas:
We also need to think about how the loop invariant optimization affects the register pressure. The main issue here is that the optimization might remove a relatively cheap calculation from the loop, but the register holding the result needs to spilled afterwards and the spilling is more expensive than the calculation itself. A good example of such a transformation is the `Brainfuck` test file.

However, at least in the case of `Brainfuck` this doesn't actually happen. As far as I can tell, the reason is that the instruction selection pulls the calculation (an addition with a constant) into the loop again in order to unite it with another expression to a single instruction.

So, the question is probably whether we think this is enough to avoid most of the possible downsides of loop invariant code motion, or whether the optimization should apply some kind of heuristic to determine whether the change is worth it or not.

Note that we can also implement such a heuristic a bit later, as reducing the number of cases where the optimization is applied should not affect its correctness.

larsk21:
Of course, I already mentioned register pressure last Monday because I was worried about the same thing.

However, for single loops, the target block (where the nodes are moved to) is usually very close to the loop entry point (in our case the header). So I don't think we loose much here. The worst case would be a node in an inner loop that is invariant to both the inner and the outer loop, and the outer loop having multiple predecessors, meaning the deepest common dominator is not directly connected to the loop entry point. I don't think this should happen too often, but I will add an entry point for a heuristic, so we can decide later if we want to avoid moving some loop-invariant nodes.

The best solution, in my opinion, would however be to have an optimization that reduces register pressure (run after the loop optimization).

larsk21:
The current strategy is to move all invariant nodes (*aggressive*). One possible alternative could be to move nodes only outside the innermost loop and only if the target block is a direct predecessor of the loop entry point (*conservative*). This would minimize register pressure as much as possible. There are, of course, also alternatives between the two extremes.

N-Maas:
Actually, my concerns are not directly about the distance of the invariant node to the loop (of course this is also relevant). The issue is that even for a simple loop where the nodes are moved to the direct predecessor block of the loop, this can in some cases significantly increase the register pressure. The pathological example would be that the loop contains _n_ nodes that are invariant. Each of the nodes has a constant as one predecessor, however the other predecessor is a node _x_ that is common to all the _n_ invariant nodes. If the invariant nodes are moved outside the loop, we now have _n_ virtual registers which are alive for the complete loop, although before it was only one.

Of course, moving invariant nodes can also decrease the register pressure, the interaction is complicated...

My conclusion is probably that the `MoveInvariantStrategy` at least needs access to the `LoopInvariantAnalysis` if we want to implement a proper heuristic later on.

N-Maas (approved):
Looks good to me apart from the open questions about the best strategy.

larsk21:
> My conclusion is probably that the MoveInvariantStrategy at least needs access to the LoopInvariantAnalysis if we want to implement a proper heuristic later on.

This can be achieved with the current implementation of the strategy pattern by simply passing the additional information required by the strategy in the constructor (as done in constant optimization, see below).

https://github.com/larsk21/compiler-minijava/blob/42395e5f470c5df35ffa2adae5d48b1689b76b78/src/main/java/edu/kit/compiler/optimizations/constant_folding/UndefinedCondStrategies.java#L50-L59

N-Maas:
> > My conclusion is probably that the MoveInvariantStrategy at least needs access to the LoopInvariantAnalysis if we want to implement a proper heuristic later on.
> 
> This can be achieved with the current implementation of the strategy pattern by simply passing the additional information required by the strategy in the constructor (as done in constant optimization, see below).
> 
> https://github.com/larsk21/compiler-minijava/blob/42395e5f470c5df35ffa2adae5d48b1689b76b78/src/main/java/edu/kit/compiler/optimizations/constant_folding/UndefinedCondStrategies.java#L50-L59

Fair :+1:

PR #146, opened 2022-01-30
Add loop unrolling optimization

g4rgoil:
This PR implements an optimization that tries to unroll loops when the number of iterations is known at compile time. Currently loops can only be unrolled by factors that evenly divide the number of iterations.


src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
```diff
@@ -0,0 +1,316 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+/**
+ * Implements an analysis to detect loops in Firm graphs.
+ */
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public final class LoopAnalysis {
```
larsk21, line 33: Do we need an additional loop analysis?

src/main/java/edu/kit/compiler/optimizations/Util.java
```diff
@@ -236,4 +238,36 @@ public void visitBlock(Block block) {
 
     }
 
+    /**
+     * Call `func` for each non-Bad direct control flow predecessor of `block`.
+     */
+    public static void forEachPredBlock(Block block, ObjIntConsumer<Block> func) {
```
larsk21, line 244: `CommonUtil.stream(block.getPreds()).filter(..).forEach(..)`?

src/main/java/edu/kit/compiler/optimizations/Util.java
```diff
@@ -236,4 +238,36 @@ public void visitBlock(Block block) {
...
+            if (pred.getOpCode() != ir_opcode.iro_Bad) {
+                func.accept((Block) pred.getBlock(), i);
+            }
+        } 
+    }
+
+    /**
+     * Returns a map containing a map of blocks to a list of contained nodes.
+     */
+    public static Map<Block, List<Node>> getNodesPerBlock(Graph graph) {
```
larsk21, line 256: `Util.BlockNodeMapper`?

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
g4rgoil: You're right that we should ideally merge the two analyses. However, given the time frame, I think it is of secondary importance. Our respective loop analyses don't necessarily compute the same set of information, so combining them would not be trivial.

src/main/java/edu/kit/compiler/optimizations/Util.java
g4rgoil: Unfortunately, I usually need both the predecessor `Block`, as well as the index of the predecessor.

src/main/java/edu/kit/compiler/optimizations/Util.java
g4rgoil: Yes, good point, I may be able to reuse that. It is not the exact same thing, the `BlockNodeMapper` includes the `Block` itself in the `List`. I'm not sure if this is relevant for my use case.

src/main/java/edu/kit/compiler/optimizations/Util.java
larsk21: Unfortunately, Java has no tuples :D

src/main/java/edu/kit/compiler/optimizations/Util.java
larsk21: Does it?

g4rgoil: Yes, `defaultVisit` is also called for blocks.

larsk21: Funny, I never realized that :D It doesn't matter for the `BlockNodeMapper` though, since `block.getBlock()` returns `null`.

src/main/java/edu/kit/compiler/optimizations/Util.java
larsk21: Not shorter or easier to read, but I'll just leave this here:

```java
public static void forEachPredBlockStream(Block block, ObjIntConsumer<Block> func) {
    IntStream.range(0, block.getPredCount())
        .mapToObj(i -> new AbstractMap.SimpleEntry<>(i, block.getPred(i)))
        .filter(item -> item.getValue().getOpCode() != ir_opcode.iro_Bad)
        .forEach(item -> func.accept((Block) item.getValue().getBlock(), item.getKey()));
}
```

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopUnroller.java
```diff
@@ -0,0 +1,295 @@
...
+            fixHeaderCond(copies.get(loop.getCond()).get(i), false);
+        }
+    }
+
+    /**
+     * Sets the control flow predecessors of the original loop. If the loops is
+     * to be fully unrolled, replaces the conditional jump in the header with
+     * an unconditional jump that exits the loop
+     */
+    private void fixOriginalHeader(Block header, NodeVec headerCopy) {
```
N-Maas, line 146: `headerCopy` is unused

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopUnrollingOptimization.java
```diff
@@ -0,0 +1,170 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+public class LoopUnrollingOptimization implements Optimization.Local {
+
+    static {
+        JFirmSingleton.initializeFirmLinux();
+    }
```
N-Maas, line 34: Personally, I would prefer a static method for `INT_MAX` instead of using such an initializer block

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopUnrollingOptimization.java
```diff
@@ -0,0 +1,170 @@
...
+    private static final Result tryUnroll(Loop loop, int iterations) {
+        var result = Result.UNCHANGED;
+        Optional<UnrollFactor> factor;
+
+        do {
+            var nodesPerBlock = Util.getNodesPerBlock(loop.getGraph());
+            factor = UnrollFactor.of(loop, iterations, nodesPerBlock);
+
+            if (factor.isPresent()) {
+                var factor_ = factor.get();
```
N-Maas, line 115: I would prefer something like `factorVal`, the `_` is easy to overlook

N-Maas (commented):
Looks good in general, I didn't find a major issue (although I didn't follow the logic in detail). There are a few places where readability could be a bit improved in my opinion, but given the limited time left it is not worth it to focus on.

However, we definitively need some tests,  primarily for the `LoopVariableAnalysis`. A number of mjtest-files with different variants of loops might be a good idea.

g4rgoil:
> Looks good in general, I didn't find a major issue (although I didn't follow the logic in detail). There are a few places where readability could be a bit improved in my opinion, but given the limited time left it is not worth it to focus on.

If you have anything you'd like me to improve in particular, feel free to let me know.

> However, we definitively need some tests, primarily for the `LoopVariableAnalysis`. A number of mjtest-files with different variants of loops might be a good idea.

Have you seen the unit tests I added, they cover a lot of the behavior of `LoopVariableAnalysis`. I'm also working on a number of `mjtest`s.

N-Maas:
> > However, we definitively need some tests, primarily for the `LoopVariableAnalysis`. A number of mjtest-files with different variants of loops might be a good idea.
> 
> Have you seen the unit tests I added, they cover a lot of the behavior of `LoopVariableAnalysis`. I'm also working on a number of `mjtest`s.

Yes, the unit tests are good for (simple) cases where the optimization should apply. But I'm more interested in cases where the optimization might be applied incorrectly, e.g. if the loop analysis thinks the variable increases constantly but it actually doesn't. For this, a few tests with loops with more complicated control flow and maybe nested loops could be a good idea.

N-Maas (approved):
Generally looks good to me. You could consider allowing a bit more than 64 nodes in case the loop can be completely unrolled (maybe depending on the number of iterations), as complete unrolling unlocks additional optimization possibilities.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
```diff
@@ -386,6 +387,7 @@ public static void main(String[] args) {
                     new ArithmeticReplacementOptimization(),
                     new LinearBlocksOptimization(),
                     new InliningOptimization(),
+                    new LoopUnrollingOptimization(),
```
larsk21, line 390: Is there a reason for the exact placement of the `LoopUnrollingOptimization` in the list of optimizations?

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopVariableAnalysis.java
```diff
@@ -0,0 +1,232 @@
...
+        return switch (node.getOpCode()) {
+            case iro_Const -> ((Const) node).getTarval();
+            default -> BAD;
+        };
+    }
+
+    /**
+     * Returns the supremum of the two Tarvals.
+     */
+    private static TargetValue supremum(TargetValue lhs, TargetValue rhs) {
```
larsk21, line 219: Just as a note: we also have the `TagetValueLatticeElement` class implementing this functionality.

src/main/java/edu/kit/compiler/JavaEasyCompiler.java
larsk21: For example, we don't have to unroll loop invariant code.

g4rgoil: I mainly considered the interactions between inlining and loop unrolling. Moving loop invariant code first should allow us to unroll loops more frequently, as it may decrease the size of loop bodies. I'll move the optimization accordingly.

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopVariableAnalysis.java
g4rgoil: To be honest, for this very limited use case, I found the extra layer of abstraction a bit inconvenient.

PR #147, opened 2022-01-30
Consider loop depth for inlining priority

N-Maas:


larsk21 (approved):
Correct usage of the loop analysis. I have no idea if the calculated weight does what it's supposed to do.

PR #151, opened 2022-01-31
Assure dominance information

larsk21:


N-Maas:
Could you also add the failing test case here?

N-Maas (approved):

PR #152, opened 2022-01-31
Fix minor issue in semantic checks

g4rgoil:
Fixes a bug, where a `NullPointerException` could occur during semantic checks, if an unresolved identifier was used in the main method of a program.

### DIFF
--- a/src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
+++ b/src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
@@ -184,8 +184,8 @@ class MethodCheckVisitor implements AstVisitor<Boolean> {
     }
 
     public Boolean visit(IdentifierExpressionNode expr) {
-        if (isMain && !expr.isHasError() && expr.getDefinition()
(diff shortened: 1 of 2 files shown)


src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
```diff
@@ -184,7 +184,8 @@ public Boolean visit(ArrayAccessExpressionNode expr) {
     }
 
     public Boolean visit(IdentifierExpressionNode expr) {
-        if (isMain && expr.getDefinition().getKind() == DefinitionKind.Parameter) {
+        if (isMain && !expr.isHasError() && expr.getDefinition()
+                .getKind() == DefinitionKind.Parameter) {
```
larsk21, line 188: Correct, but that's an interesting formatting

larsk21 (approved):

src/main/java/edu/kit/compiler/semantic/SemanticChecks.java
g4rgoil: :disappointed:

PR #153, opened 2022-02-01
Add `MinimizeRegisterPressure` strategy to `LoopInvariantOptimization`

larsk21:
+ correctly call `confirmProperties(NONE)` in `LinearBlocksOptimization`

### DIFF
--- a/src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
+++ b/src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
@@ -2,8 +2,11 @@ package edu.kit.compiler.optimizations;
 
 import static firm.bindings.binding_irgraph.ir_graph_properties_t.IR_GRAPH_PROPERTY_CONSISTENT_DOMINANCE;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


src/main/java/edu/kit/compiler/optimizations/LoopInvariantOptimization.java
```diff
@@ -2,43 +2,62 @@
...
+        Map<Block, Set<Block>> loops = loopInvariantAnalysis.getLoops();
         Map<Block, List<Node>> loopInvariantNodes = loopInvariantAnalysis.getLoopInvariantNodes();
 
-        MoveInvariantStrategy moveInvariantStrategy = new MoveInvariantStrategies.MoveAlways();
+        boolean backEdgesEnabled = BackEdges.enabled(graph);
+        if (!backEdgesEnabled) {
+            BackEdges.enable(graph);
+        }
+
+        MoveInvariantStrategy moveInvariantStrategy = this.moveInvariantStrategy.orElse(new MoveInvariantStrategies.MinimizeRegisterPressure(loops, loopInvariantNodes));
```
N-Maas, line 57: This line is a bit long^^

src/main/java/edu/kit/compiler/optimizations/NodeCosts.java
```diff
@@ -0,0 +1,70 @@
...
+            case iro_Mul:
+                return 4;
+            case iro_Mulh:
+                return 4;
+            case iro_Div:
+                return 25;
+            case iro_Mod:
+                return 25;
+            case iro_Alloc:
+                return 50;
```
N-Maas, line 48: Alloc usually involves multiple loads, and would thus probably be even more expensive

src/main/java/edu/kit/compiler/optimizations/NodeCosts.java
```diff
@@ -0,0 +1,70 @@
...
+            case iro_Load:
+                return 50;
```
N-Maas, line 50: Assuming we might have an L1-cache hit in many cases, 50 is a bit high of an estimate, but should be ok for our needs

src/main/java/edu/kit/compiler/optimizations/NodeCosts.java
```diff
@@ -0,0 +1,70 @@
...
+            case iro_Store:
+                return 50;
+            case iro_Call:
+                return 2;
```
N-Maas, line 54: Shouldn't the cost estimate for a call include the execution of the whole function behind the call (and thus be very high)?

src/main/java/edu/kit/compiler/optimizations/loop_invariant/MoveInvariantStrategies.java
```diff
@@ -19,10 +39,176 @@ private MoveInvariantStrategies() {}
...
+         * a block to the immediate dominator of the block. If the immediate
+         * dominator of a block is not one of its direct predecessors, then the
+         * step is counted twice.
+         */
+        private int moveDistance(Node node, Block target) {
+            int distance = 0;
+
+            Block block = (Block) node.getBlock();
+            // try to move to the target block, but never further
+            while (!block.equals(target) && Util.findDeepestCommonDominator(Arrays.asList(block, target)).equals(target)) {
```
N-Maas, line 180: this line is also a bit long

N-Maas (commented):
Generally, the heuristic looks good to me

src/main/java/edu/kit/compiler/optimizations/NodeCosts.java
larsk21: True, I looked it up again. Interesting how fast caches have gotten.

src/main/java/edu/kit/compiler/optimizations/NodeCosts.java
larsk21: Ideally yes, but there is no reasonable estimate for the execution time of a function. Of course, the heuristic treats function calls as expensive.

N-Maas (approved):

PR #157, opened 2022-02-03
Fix issue in handling of temporary registers

N-Maas:
Fixes a bug where the register allocator assumed a value to be present that is actually overwritten.


g4rgoil (approved):
Seems to fix the issue, `Mod` and `DivReplacment` Tests run fine now.

PR #159, opened 2022-02-03
fix: Bugs relating to const functions

g4rgoil:
This PR fixes two bugs relating to _const_ functions:

1. The inliner would not correctly copy set the `NoMem` of inlined functions.
2. The `PureFunctionAnalysis` would not terminate reasonably, because of an incorrect recursive algrithm


N-Maas (approved):
The fix for the inliner is definitively correct, and the issue with the pure function optimization seems to be fixed, too

PR #160, opened 2022-02-03
Improve handling of Phis in loop variable analysis

g4rgoil:
Previously `Phi` nodes on any path of a loop variable would always result in a non-constant step value. This PR improves the handling of `Phi`s slightly, to allow for a limited number of cases to still detect a constant step value.


larsk21 (approved):
Looks useful

PR #161, opened 2022-02-03
Fix a bug in the `PureFunctionOptimization`

g4rgoil:
Previously the `CallCollector` of the pure function optimization, would call the analysis to compute function attributes. This is problematic, as that analysis also modifies the `visited` flags of the Firm graph, which could lead to unpredictable behavior.

Note: _This is quite difficult to reproduce in practice, hence the lack of a corresponding test case._


larsk21 (approved):
Looks cleaner now

PR #164, opened 2022-02-06
Fix issues related to loop unrolling

g4rgoil:
1. Fix a simple `NullPointerException` related to an incorrect assumption about `Phi`s in the loop header.
2. Only try to unroll _closed_ loops, i.e. loops that don't contain jumps that leave the loop in their body.

These fixes fully resolve the following two issues found in the review.

> - Falsche Ausgabe mit -O1
> ```java
> class test {
>     public static void main(String[] args) {
>         System.out.println(new test().hindex());
>     }
>     
>     public int hindex() {
>         int i = 0;
>                 int[] help = new int[2];
>         help[0] = 0;
>         help[1] = 0;
>                 i=0;
>         while (i <= 1) {
>             if (i>help[i]) return i-1;
>             i=i+1;
>         }
>         return 999;
>     }       
> }
> ```

>  - (Simon_164, Simon_165, licm)
> ```
> Exception in thread "main" java.lang.NullPointerException: Cannot invoke edu.kit.compiler.optimizations.unrolling.LoopUnroller$NodeVec.get(int)" because "predCopy" is null
>         at du.kit.compiler.optimizations.unrolling.LoopUnroller.fixHeaderPhi(LoopUnroller.java:204)
>         at edu.kit.compiler.optimizations.unrolling.LoopUnroller.fixHeaderPhis(LoopUnroller.java:191)
>         at edu.kit.compiler.optimizations.unrolling.LoopUnroller.unroll(LoopUnroller.java:94)
>         at edu.kit.compiler.optimizations.unrolling.LoopUnroller.unroll(LoopUnroller.java:68)
> ```
> ```java
> int n = 3;
> int x8 = 8;
> int x9 = 9;
> while (n > 0) {
>     x9 = x8;
>     n = n - 1;
> }
> System.out.println(x9);
> ```

The warning in the following example is also resolved. However the `while(true)` loop still disappears, which does not seem related to loop unrolling.

> - (Sorter)
> ```
>    Verify warning: Cond T[546:374](Sorter_main_4[144]): Cond node lacks false proj
>    error: verifier failed; trying to write assert graph and abort
> ```
> mit --compile-firm bei
> ```java
> while (true) {
>     int j = 0;
>     while (j < 999 && j < 1000) {
>         j = j + 1;
>     }
> }
> ```

### DIFF
--- a/src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
+++ b/src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
@@ -95,6 +95,10 @@ public final class LoopAnalysis {
         return surroundingLoops;
     }
 
+    /**
+     * Returns the nearest surrounding loop of `block`, where `block` is not the
+     * header of that loop.
+     */
     private Loop findSurroundingLoop(Block block) {
         var dom = block.ptr;
         while ((dom = binding_irdom.get_Block_idom(dom)) != null) {
@@ -119,9 +123,6 @@ public final class LoopAnalysis {
                 var loop = surroundingLoops.get(block);
                 Util.forEachPredBlock(block, (predBlock, i) -> {
                     var predLoop = surroundingLoops.get(predBlock);
-                    if (predLoop == null) {
-                        return;
-                    }
                     if (predLoop != null && !predLoop.equals(loop)
                             && !predLoop.getHeader().equals(block)) {
                         // this CF edge is a jump from within a loop to a block
--- a/src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
+++ b/src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
@@ -2,8 +2,6 @@ package edu.kit.compiler.optimizations.unrolling;
 
 import static firm.bindings.binding_irgraph.ir_graph_properties_t.IR_GRAPH_PROPERTY_CONSISTENT_DOMINANCE;
 
-import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -22,6 +20,8 @@ import firm.nodes.Cond;
 import firm.nodes.Node;
 import firm.nodes.Proj;
 import lombok.AccessLevel;
+import lombok.EqualsAndHashCode;
+import lombok.EqualsAndHashCode.Include;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.ToString;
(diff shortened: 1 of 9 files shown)


src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
```diff
@@ -83,10 +71,66 @@ public void visitBlock(Block block) {
...
+                }
+            }
+        });
+
+        return surroundingLoops;
+    }
+
+    private Loop findSurroundingLoop(Block block) {
+        var dom = block.ptr;
+        while ((dom = binding_irdom.get_Block_idom(dom)) != null) {
```
N-Maas, line 100: I'm not completely sure I understand why this loops starts with the `idom` of `block` instead of `block` itself. This will cause the header to be not included, right? But wouldn't including the header simplify `checkLoopsClosed` (by not requiring the extra check for the header anymore)?

N-Maas (commented):
Generally looks good to me

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
g4rgoil: You're right that including the header would get rid of the need for that particular check. However, defining the surrounding loops as it is done here, has a number of advantages. For one thing, to walk up the hierarchy of loops, you can simply look up the surrounding loop of each loop header. It is also more consistent with the concept of a _closed_ loop, as it is perfectly fine for jump to exist from the loop header to outside the loop.

I considered both versions of the definition and found that not including the loop header to be more consistent (also intuitively speaking, a loop does not really surround its own header in my view).

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
```diff
@@ -83,10 +71,66 @@ public void visitBlock(Block block) {
...
+        graph.walkBlocks(new BlockWalker() {
+            @Override
+            public void visitBlock(Block block) {
+                var loop = surroundingLoops.get(block);
+                Util.forEachPredBlock(block, (predBlock, i) -> {
+                    var predLoop = surroundingLoops.get(predBlock);
+                    if (predLoop == null) {
+                        return;
+                    }
+                    if (predLoop != null && !predLoop.equals(loop)
```
0xNB, line 125: Am I misunderstanding something or do you check if the predecessor of a block is not contained inside another loop that is not its own surrounding loop? How does this handle jumps to other blocks that are not contained in any loops themselves?

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
N-Maas: Okay, then you could maybe add a comment that explains that the header is intentionally exluded

src/main/java/edu/kit/compiler/optimizations/unrolling/LoopAnalysis.java
g4rgoil: This checks, whether the control flow edge is a jump from within the body of a loop to a block that is not part of that loop. So, whether `predblock` is inside the body of `predLoop`,  `block` is inside the body of `loop`, `loop` and `predLoop` are not the same loop, and whether the jump is not from body to header.

Also, the second `predLoop != null` is a bit ugly.

0xNB: Okay but what happens if you have a jump to another block that is not contained within a loop? Then`predLoop` should also be null since no predLoop exists right?

g4rgoil: First of all, `pred` is the jump source, not the destination (reverse control flow dependency). Only the one very specific case is of interest here. I don't need to care about jumps into a loop, or jumps where no loop is involved at all.

N-Maas (approved):

PR #165, opened 2022-02-06
Minor fix in name/type analysis

g4rgoil:


larsk21:
Thanks for spotting that error! However, I think it's better to fix this a bit upstream where the data type is extracted from the definition (see #168).

PR #167, opened 2022-02-06
Fix a regression in the pure function optimization

g4rgoil:
This bug was introduced as part of #161. The fix prevents the `while (true)` from _disappearing_ in the following snippet.

> ```java
> while (true) {
>     int j = 0;
>     while (j < 999 && j < 1000) {
>         j = j + 1;
>     }
> }
> ```

### DIFF
--- a/src/main/java/edu/kit/compiler/optimizations/PureFunctionOptimization.java
+++ b/src/main/java/edu/kit/compiler/optimizations/PureFunctionOptimization.java
@@ -26,13 +26,15 @@ public final class PureFunctionOptimization implements Optimization.Local {
         for (var call : collector.calls) {
             var attributes = analysis.getAttributes(Util.getCallee(call));
             
-            hasChanged |= attributes.isPure() && unpinCall(call);
+            if (attributes.isTerminates()) {
+                hasChanged |= attributes.isPure() && unpinCall(call);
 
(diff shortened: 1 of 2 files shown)


src/main/java/edu/kit/compiler/optimizations/PureFunctionOptimization.java
```diff
@@ -26,13 +26,15 @@ public boolean optimize(Graph graph, OptimizationState state) {
...
-            };
+            if (attributes.isTerminates()) {
+                hasChanged |= attributes.isPure() && unpinCall(call);
+
+                hasChanged |= switch (attributes.getPurity()) {
+                    case CONST -> handleConstCall(call);
+                    case PURE -> handlePureCall(call, collector.usedCalls); 
+                    default -> false;
+                };
+            }
```
larsk21, line 37: So we still flag non-terminating functions as pure/const, but we don't change the pin state / memory chain, right?

g4rgoil: Yes, that is the idea. The `purity` and `terminates` attributes are orthogonal to each other. Otherwise, we would introduce a bunch of illegal state into the attributes.

N-Maas (approved):
Looks good to me

PR #168, opened 2022-02-06
make access to definition type safer in detailed type analysis

larsk21:


g4rgoil (commented):
Unfortunately I'm getting unit test failures, if I try to build this.

But generally, your approach is probably better than #165. I'll go ahead and close that one, so you can change this PR to `master` and get CI runs.

larsk21:
> Unfortunately I'm getting unit test failures if I try to build this.

Yeah, I probably should've tested it before I pushed it :D

g4rgoil (approved):
> Yeah, I probably should've tested it before I pushed it :D

It happens X).

PR #170, opened 2022-02-07
Preparation for final submission

N-Maas:
Maven doesn't seem to have support for lock files (https://stackoverflow.com/questions/44521542/why-doesnt-gradle-or-maven-have-a-dependency-version-lock-file), so the runtime dependencies are vendored.


larsk21:
Submitted successfully, but there is no need to merge this in `master`.

PR #171, opened 2022-02-07
Fix endless recursion in unused args analysis

g4rgoil:
`UnusedArgumentsAnalysis` would run into a stack overflow, if an argument to a non-inlined function was part of a loop in that functions graph.


N-Maas (approved):
Whoops :sweat_smile:

